PR #3, opened 2022-07-15
endpoint_route_handler: fix cross worker lookup

simahawk:
TODO
- [x] add mig step for endpoint and shopfloor app (PR to come soon)
- [ ] update roadmap

Companion PRs:
- shopfloor https://github.com/OCA/wms/pull/574
- edi https://github.com/OCA/edi/pull/731


endpoint/views/endpoint_view.xml
```diff
@@ -10,13 +10,26 @@
...
+                    <div
+                        class="alert alert-warning"
+                        role="alert"
+                        attrs="{'invisible': [('registry_sync', '=', True)]}"
+                    >
+                        <p>
+                            Registry out of sync. Likely the record has been modified but not sync'ed with the routing registry.
+                        </p>
+                        <p>
+                            Use the action "Sync route registry" to make changes effective once you are done with edits and creates.
```
sebalix:
```suggestion
                            Use the action "Sync registry" to make changes effective once you are done with edits and creates.
```

endpoint_route_handler/registry.py
```diff
@@ -2,7 +2,58 @@
...
+
+from odoo import http, tools
+from odoo.tools import DotDict
+
+from odoo.addons.base.models.ir_model import query_insert
+
+from .exceptions import EndpointHandlerNotFound
+
+
+def query_multi_update(cr, table_name, rows, cols):
```
mmequignon, line 24: Why is it a function instead of a method ? cr and table_name are in self, it would have made sense to me.

endpoint_route_handler/registry.py
```diff
@@ -2,7 +2,58 @@
...
+        keys=keys,
+        col_names=col_names,
+        values=sql.Placeholder(),
+    )
+    execute_values(
+        cr,
+        query.as_string(cr._cnx),
+        rows,
+        template=template.as_string(cr._cnx),
+    )
```
mmequignon, line 60: Can't you pass a dict here, and do it all at once, instead of format then execute_values?
I think this could be improved :blush:

endpoint_route_handler/registry.py
```diff
@@ -11,103 +62,232 @@ class EndpointRegistry:
...
+            yield EndpointRule.from_row(self.cr.dbname, row)
 
-    def get_rules(self):
-        return self._mapping.values()
+    def _get_rules(self, keys=None, where=None):
+        query = "SELECT * FROM endpoint_route"
+        if keys and not where:
+            where = "key in (%s)"
+            self.cr.execute(query, keys)
+            return self.cr.fetchall()
```
mmequignon: The where variable isn't used there, and the two last lines looks redundant.

endpoint_route_handler/registry.py
```diff
@@ -11,103 +62,232 @@ class EndpointRegistry:
...
+        elif where:
+            query += " " + where
+        self.cr.execute(query)
+        return self.cr.fetchall()
+
+    def _get_rule(self, key):
+        query = "SELECT * FROM endpoint_route WHERE key = %s"
+        self.cr.execute(query, (key,))
+        row = self.cr.fetchone()
```
mmequignon: I might have missed it : Do we have a unique constraint on the key column?

endpoint_route_handler/registry.py
```diff
@@ -11,103 +62,232 @@ class EndpointRegistry:
...
+    def drop_rules(self, keys):
+        self.cr.execute("DELETE FROM endpoint_route WHERE key IN %s", (tuple(keys),))
+        return True
 
     @classmethod
-    def registry_for(cls, dbname):
-        if dbname not in _REGISTRY_BY_DB:
-            _REGISTRY_BY_DB[dbname] = cls()
-        return _REGISTRY_BY_DB[dbname]
+    def registry_for(cls, cr):
```
mmequignon: I don't want to be picky here, but aren't class methods supposed to be placed above regular methods ?

endpoint_route_handler/models/endpoint_route_handler.py
```diff
@@ -323,3 +305,35 @@ def _get_routing_info(self):
             csrf=self.csrf,
         )
         return route, routing, self.endpoint_hash
+
+    def write(self, vals):
+        if any([x in vals for x in self._controller_fields() + ("active",)]):
```
mmequignon: Why isn't "active" returned by `_controller_fields` ? This is a field that requires a sync, right ?

endpoint_route_handler/registry.py
simahawk: Because I could reuse it somewhere else. In fact, I should move it to `utils.py`.

endpoint_route_handler/registry.py
simahawk: dict? to what? `execute_values` wants the strings

endpoint_route_handler/registry.py
simahawk: hmm the where must be used, the return is wrong, likely a typo. Thanks!

endpoint_route_handler/registry.py
simahawk: there are comments at the top of the file: is a TODO

endpoint_route_handler/registry.py
simahawk: depends. I prefer to keep more relevant methods on the top but I'll sort them out

endpoint_route_handler/registry.py
mmequignon: it accepts dicts, as explained in the doc https://www.psycopg.org/docs/extras.html#psycopg2.extras.execute_values

simahawk: `rows` ARE dicts. I still don't get what you want to pass as dict :man_shrugging:

mmequignon: It's just a thought. Is it possible to skip the formatting, by just passing additional args to `execute_values` ?

simahawk: so: no :P

endpoint_route_handler/models/endpoint_route_handler.py
simahawk: those fields are only the ones that are used for the real route and to generate the hash. `active` shouldn't fit

simahawk: I already renamed it to `_routing_impacting_fields` anyway ;P

simahawk:
Quick notes from call w/ @lmignon 

Future improvements (out of the scope of this PR for now, if not for small changes):

* make route a multiple value (support more routes and req methods)

1. keep the same in the ui for now, later own we can imagine a multi-value selection or just add text field w/ proper validation and cleanup

* remove the route field in the table of endpoint_route
 
1. support a comma separated list of routes
2. maybe support comma separated list of methods
3. use only routing.routes for generating the rule
4. sort and freeze its values to update the endpoint hash

* catch dup route exception on the sync to detect duplicated routes
  and use the endpoint_hash to retrieve the real record
  (note: we could store more info in the routing information which will stay in the map)

* for customizing the rule behavior the endpoint the hook is to override the registry lookup

  make EndpointRule class overridable on the registry
  
  
NOTE in v16 we won't care anymore about odoo controller so the lookup of the controller can ne simplified to an basic py obj the holds the routing info

~For the main issue of "non initialize-able abstract model" reported on #12 we said that we should get rid of endpoint.route.handler features relying on `new()` and move the computation and validation of fields to a pure python object.~

simahawk:
Note to myself: we might have a workaround for v15 and above.
We could keep endpoint.route.handler as it is and keep using it as a mixin.
For the "tool" approach, in order to be able to initialize it with `new` we could have a transient model endpoint.route.handler.tool which we can use in the same fashion and that won't be created in its table at any time.
We could even use `_auto = False` to not let it create any table. 
To be tested.

Thanks @sebalix for the suggestion ;)

simahawk:
@lmignon I finally fixed the problem here and also found a workaround for v > 15 that allows to not change design with this commit [endpoint_route_handler: add new tool model](https://github.com/OCA/web-api/pull/3/commits/de04ac38e935d8956b8dc72ea9975bd9189e9260)
You can see it here on v16 too https://github.com/OCA/web-api/pull/17

endpoint_route_handler/models/endpoint_route_sync_mixin.py
```diff
@@ -116,3 +118,20 @@ def _prepare_endpoint_rules(self, options=None):
...
+def is_registry_sync_column_ready(cr, table):
+    if tools.sql.column_exists(cr, table, "registry_sync"):
+        return True
+    # Depending on your modules inheritance and upgrade order
+    # when you introduce this mixin on an existing model
+    # it might happen that `_register_hook`
+    # gets called before the model's table is ready
+    # (eg: another odoo service loading the env before the upgrade happens).
+    # Let if fail gracefully since the hook will be called again later.
+    _logger.debug(
```
sebalix: debug or warning?

sebalix (approved):

endpoint_route_handler/models/endpoint_route_sync_mixin.py
simahawk: I'm not sure this is going to be a problem as it will run whenever the env is loaded

endpoint_route_handler/registry.py
```diff
@@ -11,103 +66,330 @@ class EndpointRegistry:
...
-    @classmethod
-    def wipe_registry_for(cls, dbname):
-        if dbname in _REGISTRY_BY_DB:
-            del _REGISTRY_BY_DB[dbname]
+    def _lock_rows(self, keys):
+        sql = "SELECT id FROM endpoint_route WHERE key IN %s FOR UPDATE"
+        self.cr.execute(sql, (tuple(keys),), log_exceptions=False)
+
+    def _update(self, rows_mapping):
+        self._lock_rows(tuple(rows_mapping.keys()))
```
sebalix, line 200: Feedback: locking rows here is triggering a `psycopg2.extensions.TransactionRollbackError: could not serialize access due to concurrent update` error. Traceback is the following:
```python
[...]
  File "/odoo/src/odoo/__init__.py", line 104, in registry
    return modules.registry.Registry(database_name)
  File "/odoo/src/odoo/modules/registry.py", line 65, in __new__
    return cls.new(db_name)
  File "/odoo/src/odoo/modules/registry.py", line 89, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/odoo/src/odoo/modules/loading.py", line 571, in load_modules
    model._register_hook()
  File "/odoo/external-src/web-api/endpoint_route_handler/models/endpoint_route_sync_mixin.py", line 91, in _register_hook
    self.search(domain)._register_controllers(init=True)
  File "/odoo/external-src/wms/shopfloor_base/models/shopfloor_app.py", line 165, in _register_controllers
    super()._register_controllers(init=init, options=options)
  File "/odoo/external-src/web-api/endpoint_route_handler/models/endpoint_route_sync_mixin.py", line 102, in _register_controllers
    self._endpoint_registry.update_rules(rules, init=init)
  File "/odoo/external-src/web-api/endpoint_route_handler/registry.py", line 236, in update_rules
    self._update(to_update)
  File "/odoo/external-src/web-api/endpoint_route_handler/registry.py", line 200, in _update
    self._lock_rows(tuple(rows_mapping.keys()))
  File "/odoo/external-src/web-api/endpoint_route_handler/registry.py", line 197, in _lock_rows
    self.cr.execute(sql, (tuple(keys),), log_exceptions=False)
  File "<decorator-gen-3>", line 2, in execute
    
  File "/odoo/src/odoo/sql_db.py", line 101, in check
    return f(self, *args, **kwargs)
  File "/odoo/src/odoo/sql_db.py", line 298, in execute
    res = self._obj.execute(query, params)
psycopg2.extensions.TransactionRollbackError: could not serialize access due to concurrent update
```
This is happening when we start multiple Odoo instances at the same time (in our case it was 6), so each of them is trying to lock the same rows, generating this concurrency issue. Odoo is trying again & again to start until it successfully obtain the locks to do its stuff, and this is taking more and more time to boot depending on the number of instances to start.

We should try with a `pg_try_advisory_lock` or maybe (but I don't know the whole logic of the current code so tell me if it's possible):

- use of `FOR UPDATE SKIP LOCKED`, but we must return the ID of the rows which have been locked by doing so (`_lock_rows` has to returns these IDs): https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/
- update only these IDs in the `query_multi_update` call. Lines which have been locked by another transaction will do the same, and if a transaction has no row to update it'll do nothing.

Now if the registry loading of an Odoo instance does nothing, is it an issue for this feature? Does it need to read this table afterwards (in the same transaction) to do some stuff?

lmignon: IMO the loading of the registry should not modify the database. Only the update, install of new addon and regular operations on Odoo models should. :thinking:

simahawk: The table must be readable as soon as the routing map is generated. This happens as the 1st http request comes in. Hence, is pretty critical.

In general, I tend to say we should lock only the ones that are updated and in theory this is what should happen already.

Question: which records is trying to update? If you get this, I guess all. Which seems wrong.
By looking again at the hook, I'm not sure anymore that we should still call `register/unregister_controllers`.

The purpose of this was to make sure that all routes are correctly registered when the instance boots BUT since all routes are now stored in the DB the normal sync mechanism should take care of this when the instance is up and running.

Meaning: if you modify a route to be active/inactive, you should explicitly run the sync of the registry to register/unregister routes. There shouldn't be any reason to ensure again that they are all registered when it starts up.

And in this PR and in related PRs there's already an upgrade step that should register all existing route-dependent records. Sounds like a left-over logic from prev implementation.

I hope is clear :)

simahawk: I'm back (out of sickness finally :mechanical_arm:).
I've managed to reproduce in a dev env too.
Pushed a fix. I'll fwd port to it v16 later.

simahawk: I'm back (out of sickness finally :mechanical_arm:).
I've managed to reproduce in a dev env too.
Pushed a fix. I'll fwd port to it v16 later.

simahawk:
/ocabot merge nobump

:rocket:

simahawk:
> /ocabot merge nobump
> 
> rocket

killed the build as I spotted a typo on test class decorator

PR #8, opened 2022-09-01
[14.0] webservice: add api key and public auth support

simahawk:
ref: cos-3712

### DIFF
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -1,7 +1,9 @@
 # Copyright 2020 Creu Blanca
+# Copyright 2022 Camptocamp SA
+# @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import _, api, exceptions, fields, models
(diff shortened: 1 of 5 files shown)


webservice/models/webservice_backend.py
```diff
@@ -14,8 +16,19 @@ class WebserviceBackend(models.Model):
...
+            ("none", "Public"),
+            ("user_pwd", "Username & password"),
+            ("api_key", "API Key"),
+        ],
+        default="user_pwd",
+        required=True,
+    )
+    username = fields.Char(auth_type="user_pwd")
+    password = fields.Char(auth_type="user_pwd")
+    api_key = fields.Char(string="API Key", auth_type="api_key")
```
TDu, line 30: I don't understand how this `auth_type` parameter on the filed declaration works and where it comes from ?

simahawk: I'm adding it. See `_valid_field_parameter` override

TDu: Thanks, interesting.

TDu (approved):

etobella (approved):
Makes sense to me

PR #9, opened 2022-09-28
[15.0][MIG] endpoint_route_handler

flachica:


flachica:
@simahawk I don't find Issue for migrating PR's

flachica:
I can't solve errors in tests @simahawk. I think that is related to default values when mockup endpoints but for register and unregister service `id` is needed and the record is in memory and not has id attribute

PR #13, opened 2022-10-13
[14.0¯[FIX] enpoint_route_handler: Put in place the machinery to fix the re…

lmignon:
…load of routes in all the workers

### DIFF
--- a/endpoint/models/endpoint_mixin.py
+++ b/endpoint/models/endpoint_mixin.py
@@ -2,6 +2,7 @@
 # @author: Simone Orsi <simone.orsi@camptocamp.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+import logging
 import textwrap
 from functools import partial
 
@@ -19,7 +20,6 @@ from ..controllers.main import EndpointController
 class EndpointMixin(models.AbstractModel):
 
     _name = "endpoint.mixin"
-    _inherit = "endpoint.route.handler"
     _description = "Endpoint mixin"
 
     exec_mode = fields.Selection(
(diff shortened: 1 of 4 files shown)


lmignon:
@simahawk Here it's a working draft of a mechanism to refresh the routing map in all the workers.

In short:
* The `endpoint_route_handler` is now a concrete model so we've a way to look for all the route into a single table
* `endpoint_route_handler` should only define fields involved into the routing definition (no name, route_group, request_content_type, endpoint_hash)...
* On each create / write, the version field is automatically incremented by a DB trigger. The `check_signaling`use this value to know if the routing map must be updated 
* At this stage the `unlink` of a route handler is not ?yet? supported
* Since the trigger of the reloading is now based on a change into the endpoint version sequence, we must remove all the code checking in which case the routing map should be updated, but also the code triggering the update on create and write (the check_signaling that will occur on the next call to the worker will do the job).
* since the `endpoint.route.handler` is now a model, the use of endpoint_hash seems no more relevant and could be replaced by the id field.
* The use of `endpoint.route.handler` as a tool is no more possible. The registry trigger expect to be able to get the handler method from the concrete endpoint linked to the handler.  We could ignore the registration of an endpoint route handler into the registry if handler_model becomes optional and and not filled at creation.
....

lmignon:
@simahawk Do you plan to work on this in a near future?

Regards,

lmi

simahawk:
@lmignon ciao, actually I worked on it on my way back from BRU  but I never found the time to finalize everything and get back to you. AFAIR I took this decision: I'd like to go w/ such heavy refactoring only on 16.0 and keep my approach + timestamp for modification and route update on v14 (what I started on #3). I'll finalize v14 soon.

There are many details of the big refactoring I'd like to discuss w/ you, maybe we can organize a call before the end of the year or in Jan to check together.
Hope is ok for you :)

lmignon:
Hi @simahawk 

>  I'd like to go w/ such heavy refactoring only on 16.0 

For sure, it's safer and logic.
 
> There are many details of the big refactoring I'd like to discuss w/ you, maybe we can organize a call before the end of the year or in Jan to check together.
Hope is ok for you :)

It's ok for me. My main concern is to move forward with all this as this addon is increasingly on my critical path to finalize fastapi as it should be used in production by the end of January. In addition, I will be away from 23 December to 8 January.

Regards,

lmi

lmignon:
superseded by #17

PR #17, opened 2023-02-17
[16.0] endpoint_route_handler: migration to v16

simahawk:
Includes #3


endpoint_route_handler/models/endpoint_route_sync_mixin.py
```diff
@@ -0,0 +1,118 @@
...
+    registry_sync = fields.Boolean(
+        help="ON: the record has been modified and registry was not notified."
+        "\nNo change will be active until this flag is set to false via proper action."
+        "\n\nOFF: record in line with the registry, nothing to do.",
+        default=False,
+        copy=False,
+    )
+
+    def write(self, vals):
+        if any([x in vals for x in self._routing_impacting_fields() + ("active",)]):
```
lmignon, line 43: `_routing_impacting_fields`is not defined here and should be mentioned to be implemented into consumers of this mixin

lmignon (approved):
Thank you @simahawk Works like a charm. I refactored the code of [odoo-addon-fastapi](https://github.com/OCA/rest-framework/pull/291/commits/e510d123acdc451a3e5f7cee9ea7a0fbc347f402) to use your new mixin and everything is smooth.

endpoint_route_handler/models/endpoint_route_sync_mixin.py
simahawk: I added it on v14 and not ported the fix here. I'll do it.
Thanks for testing :)

endpoint_route_handler/models/ir_http.py
```diff
@@ -0,0 +1,82 @@
...
+        e_registry = cls._endpoint_route_registry(http.request.env)
+        for endpoint_rule in e_registry.get_rules():
+            _logger.debug("LOADING %s", endpoint_rule)
+            endpoint = endpoint_rule.endpoint
+            for url in endpoint_rule.routing["routes"]:
+                yield (url, endpoint)
+
+    @classmethod
+    def routing_map(cls, key=None):
+        last_update = cls._get_routing_map_last_update(http.request.env)
```
lmignon: @simahawk To be efficient, we should not rely on the last_update but on the last value of a dedicated sequence. The computation of the last_update requires a query on the whole table followed by a sort on the last_update column. This kind of operation is far more resources consuming than a simple query to get the `last_value` from a sequence.

endpoint_route_handler/registry.py
```diff
@@ -0,0 +1,342 @@
...
+
+            cr.execute(
+                """
+                CREATE OR REPLACE FUNCTION endpoint_route_set_timestamp()
+                    RETURNS TRIGGER AS $$
+                BEGIN
+                    NEW.updated_at = NOW();
+                    RETURN NEW;
+                END;
+                $$ LANGUAGE plpgsql;
```
lmignon: Here we should create a sequence `endpoint_version` and add a trigger to auto increment the sequence on update, insert and delete.

endpoint_route_handler/registry.py
```diff
@@ -0,0 +1,342 @@
...
+        return res
+
+    def drop_rules(self, keys):
+        self.cr.execute("DELETE FROM endpoint_route WHERE key IN %s", (tuple(keys),))
+        return True
+
+    def make_rule(self, *a, **kw):
+        return EndpointRule(self.cr.dbname, *a, **kw)
+
+    def last_update(self):
```
lmignon, line 247: should be replaced by `def last_version(self)`

```suggestion
    def last_version(self):
          self.cr.execute("SELECT last_value FROM endpoint_version")
          return self.cr.fetchone()[0]
```

lmignon (changes requested):

flachica (approved):
Functional testing with oca/rest-framework/fastapi is ok

endpoint_route_handler/registry.py
simahawk: I introduced a sequence on v14 here https://github.com/OCA/web-api/pull/3/commits/3ae6ceb6ef6ba83682b5d9037792cf5d3fba178d

I'm testing. I'll fwd port soon.

simahawk:
All changes from v14 ported, including the use of sequence instead of timestamp to detect updates :)

lmignon:
@simahawk When can we expect to have this one merged? It's blocking the `fastapi` addon and I would like to publish a first release of this since Odoo 16 is out from a long time now...

simahawk:
> @simahawk When can we expect to have this one merged? It's blocking the `fastapi` addon and I would like to publish a first release of this since Odoo 16 is out from a long time now...

I know... In the end I'm happy I kept it on hold because there was a bug that I fixed on v14 which was breaking SF.
I'll merge 14 soon and port the latest version here.

simahawk:
@lmignon I've done a "blind" porting of the latest version from 14.0 here.
I don't have time to test but I wanted to unlock your work if possible.
Can you please take care of testing and reviewing again pls?

xAdrianCif (approved):
Implementation test and code review OK

lmignon:
@simahawk Thank you for the last changes. At this stage I still encounter an issue when I sync the registry. I need to restart my instance to make the  route available....  It seems it only occurs for endpoints created from a xml record the first time it's activated...  It not occurs in others cases (If I add a new endpoint, our update endpoints created from xml record after restart ...)

Not a blocking issue.

lmignon (approved):

simahawk:
> @simahawk Thank you for the last changes. At this stage I still encounter an issue when I sync the registry. I need to restart my instance to make the route available.... It seems it only occurs for endpoints created from a xml record the first time it's activated... It not occurs in others cases (If I add a new endpoint, our update endpoints created from xml record after restart ...)
> 
> Not a blocking issue.

Thanks for your feedback!
In fact, this is by design. 
IMO you should enable endpoints explicitly. I'm open for discussion of course. Feel free to open an issue :wink: 

/ocabot merge nobump

PR #21, opened 2023-06-09
[16.0][MIG] endpoint

JordiBForgeFlow:
Standard migration


JordiBForgeFlow:
cc @EvaSForgeFlow

OriolMForgeFlow (approved):
LGTM!

endpoint/hooks.py
```diff
@@ -1,76 +0,0 @@
...
-
-_logger = logging.getLogger(__file__)
-
-
-def post_init_hook(cr, registry):
-    env = api.Environment(cr, SUPERUSER_ID, {})
-    _init_server_action(env)
-
-
-def _init_server_action(env):
```
simahawk: why trashing this?

simahawk (changes requested):

endpoint/hooks.py
OriolMForgeFlow: In version 16, the field 'activity_user_type' is no longer required=True. It has been changed to be computed. 
As a result, the error `psycopg2.IntegrityError: null value in column "activity_user_type" violates the not-null constraint`, which was previously fixed by the hook in previous versions, is no longer appearing.

simahawk: Ok, but where is the server action then? :)

OriolMForgeFlow: You are right, now it is loaded from a data file.

simahawk (approved):

PR #22, opened 2023-06-22
[14.0] webservice: add webserviceConsumerMixin

petrus-v:
This is part of what has been discussed here: https://github.com/OCA/edi/pull/776#pullrequestreview-1462480583

In draft at the moment, I'd like to 
- [x] Improve and discuss about design: I felt some abstraction is missing... I'll be back on it later
- [x] add test on the mixin

### DIFF
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -2,8 +2,11 @@
 # Copyright 2022 Camptocamp SA
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from contextlib import contextmanager
 
-from odoo import _, api, exceptions, fields, models
+from requests import HTTPError
+
+from odoo import _, api, exceptions, fields, models, registry
 
 
 class WebserviceBackend(models.Model):
@@ -36,6 +39,13 @@ class WebserviceBackend(models.Model):
             ("application/x-www-form-urlencoded", "Form"),
         ],
         required=True,
+    )
+    save_response = fields.Boolean(
+        default=False,
+        help=(
+            "If enabled, the response for the external call will be saved on the "
+            "webservice consumer record",
+        ),
     )
 
     @api.constrains("auth_type")
@@ -71,8 +81,38 @@ class WebserviceBackend(models.Model):
         extra_params = ("auth_type",)
         return name in extra_params or super()._valid_field_parameter(field, name)
 
-    def call(self, method, *args, **kwargs):
-        return getattr(self._get_adapter(), method)(*args, **kwargs)
+    @contextmanager
+    def _consumer_record_env(self, record, new_cursor=False):
+        """Return the current record in a new transaction or in the the current transaction"""
+        if new_cursor:
+            with api.Environment.manage():
+                with registry(self.env.cr.dbname).cursor() as new_cr:
+                    new_env = api.Environment(new_cr, self.env.uid, self.env.context)
+                    # in case of error the main transaction will be rollback
+                    # in any case we want to save the response payload for
+                    # later analysis. Some web-services gives information regarding
+                    # the cause of failures
+                    yield record.with_env(new_env)
+        else:
+            yield record
+
+    def call(self, method, *args, consumer_record=None, **kwargs):
+        content = False
+        status_code = False
+        try:
+            content = getattr(self._get_adapter(), method)(*args, **kwargs)
+            status_code = 200
+        except HTTPError as request_error:
+            content = request_error.response.content
+            status_code = request_error.response.status_code
+            raise request_error from request_error
+        finally:
+            if self.save_response and consumer_record:
+                with self._consumer_record_env(
+                    consumer_record, new_cursor=status_code != 200
+                ) as consumer_record_tx:
+                    consumer_record_tx._save_ws_response(content, status_code)
+        return content
 
     def _get_adapter(self):
         with self.work_on(self._name) as work:
--- a/webservice/models/webservice_consumer_mixin.py
+++ b/webservice/models/webservice_consumer_mixin.py
@@ -1,0 +1,62 @@
+# Copyright 2022 Foodles (http://www.foodles.co).
+# @author Pierre Verkest <pierreverkest84@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import base64
+
+from odoo import fields, models
+from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT as DATETIME_FORMAT
+
+from odoo.addons.http_routing.models.ir_http import slugify
+
+
+class WebserviceConsumerMixin(models.AbstractModel):
+
+    _name = "webservice.consumer.mixin"
+    _description = "Add fields to save web service responses"
+
+    ws_response_status_code = fields.Integer(
+        "Status code",
+        help="Web service response HTTP code",
+    )
+    ws_response_content = fields.Binary(
+        "Response",
+        attachment=True,
+        copy=False,
+        help="Web service response content",
+    )
+    ws_response_date = fields.Datetime(
+        "Response date",
+        help="Date when the web service response has been saved",
+    )
+    ws_response_content_filename = fields.Char(
+        compute="_compute_ws_response_content_filename"
+    )
+
+    def _compute_ws_response_content_filename(self):
+        for rec in self:
+            if rec.ws_response_date and rec.ws_response_status_code:
+                formatted_response_date = slugify(
+                    rec.ws_response_date.strftime(DATETIME_FORMAT)
+                )
+                rec.ws_response_content_filename = "response_%s_%s.json" % (
+                    formatted_response_date,
+                    str(rec.ws_response_status_code),
+                )
+            else:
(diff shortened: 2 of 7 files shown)


webservice/components/request_adapter.py
```diff
@@ -16,13 +20,26 @@ class BaseRestRequestsAdapter(Component):
     _webservice_protocol = "http"
 
     # TODO: url and url_params could come from work_ctx
-    def _request(self, method, url=None, url_params=None, **kwargs):
+    def _request(
+        self, method, url=None, url_params=None, consumer_record=None, **kwargs
+    ):
+        if self.collection.save_response and isinstance(
```
simahawk: You should wrap the whole existing code of this method w/ ctx manager to:
1. check this only once
2. store the reponse content even when `request.raise_for_status` at the end will brake the transaction

Also, why do you need this `consumer_record`?
The adapter should be agnostic and know nothing about such records.
I would wrap `call` instead and make it accept this param. 
You can then define a method on the backend `_must_save_response` which checks the `save_response` flag and the consumer record if given.

webservice/models/webservice_consumer_mixin.py
```diff
@@ -0,0 +1,28 @@
...
+        "Response", attachment=True, copy=False, help="Web service response content"
+    )
+    ws_response_content_filename = fields.Char(
+        compute="_compute_ws_response_content_filename"
+    )
+
+    def _compute_ws_response_content_filename(self):
+        for rec in self:
+            rec.ws_response_content_filename = (
+                "response_%s.json" % (str(rec.ws_response_status_code),)
```
simahawk: I would add the formatted datetime string too

simahawk (changes requested):

webservice/models/webservice_consumer_mixin.py
petrus-v: I wanted to reuse efficiently all the business rules implemented here https://github.com/OCA/edi/blob/14.0/edi_oca/models/edi_exchange_type.py#L213

petrus-v: at the moment this is a computed field not stored :thinking: I could add a response_at datetime field :thinking:

webservice/components/request_adapter.py
petrus-v: Thanks for your review and feedback. I've refactor the code with this commit: 88621700be1f252c9fde80055fdb28ead439dac0 does it looks better ?

petrus-v: hum... in fact while using with [#oca/edi/787](https://github.com/OCA/edi/pull/787) I'm getting something weird as the main transction try to write on the same record as the new transcation to save the result, in the test I've done only the result paylaod was saved but the state of the `edi.exchange.record` was `output_pending` (I expected `output_sent`) :thinking:

petrus-v: logs confirmed that hypotesis

```
ERROR ** odoo.sql_db: bad query: UPDATE "edi_exchange_record" SET "edi_exchange_state"='output_sent',"exchanged_on"='2023-09-05 12:59:11',"write_uid"=62,"write_date"=(now() at time zone 'UTC') WH│
│ERROR: could not serialize access due to concurrent update
```

Not sure what's the best to do :thinking:

webservice/models/webservice_backend.py
```diff
@@ -37,6 +41,13 @@ class WebserviceBackend(models.Model):
         ],
         required=True,
     )
+    save_response = fields.Boolean(
+        default=False,
+        help=(
+            "Control whatever the webservice response should be saved on the "
+            "webservice consumer record if defined while processing the HTTP end point",
```
simahawk:
```suggestion
            "If enabled, the response for the external call will be saved on the "
            "webservice consumer record",
```

webservice/models/webservice_consumer_mixin.py
simahawk: An alternative would be to relate the ir.attachment and read the write date from there...
No strong opinion on this but I think it would be nice to know when the response was recorded.

webservice/models/webservice_backend.py
```diff
@@ -71,8 +82,40 @@ def _valid_field_parameter(self, field, name):
         extra_params = ("auth_type",)
         return name in extra_params or super()._valid_field_parameter(field, name)
 
-    def call(self, method, *args, **kwargs):
-        return getattr(self._get_adapter(), method)(*args, **kwargs)
+    @contextmanager
+    def _consumer_record_env(self, record, new_cursor=True):
```
simahawk: I'm not sure to understand why we need `new_cursor` flag. See below.

webservice/models/webservice_backend.py
```diff
@@ -71,8 +82,40 @@ def _valid_field_parameter(self, field, name):
...
+        try:
+            content = getattr(self._get_adapter(), method)(*args, **kwargs)
+            status_code = 200
+        except HTTPError as request_error:
+            content = request_error.response.content
+            status_code = request_error.response.status_code
+            raise request_error from request_error
+        finally:
+            if self.save_response and consumer_record:
+                with self._consumer_record_env(
```
simahawk, line 111: IMO the new_cursor flag shouldn't be passed to `call`, it should be _determined_ by call.
If you catch the exception you can set the flag to true and pass it to the ctx manager.
Also, when you get an HTTPError you raise immediately instead of storing that response too. 
This would solve the problem no?

webservice/models/webservice_consumer_mixin.py
petrus-v: indeed it's a nice to have !

webservice/models/webservice_backend.py
petrus-v: if I create a new transaction only in case of failure that will definitively works... expect if code is raising later in the stack the response won't be saved, this is kind of okay in my use case.

Saving response only as ir.attachement would solve the problem too but without back ref on the consumer records it will be hard to reuse the response later so I'll go for the first suggestion, creating new transaction only in case of failure !

simahawk: I see your point but probably this way is more consistent?
As an alternative we could use a savepoint later...

webservice/models/webservice_consumer_mixin.py
petrus-v: done by https://github.com/OCA/web-api/pull/22/commits/b5f821546450888863073d38cc74f02fbffbcab4

webservice/models/webservice_backend.py
petrus-v: For the time being, as I've force `status_code = 200` I'm finally creating new cursor only in case of failures (`status_code != 200`)  with this commit https://github.com/OCA/web-api/pull/22/commits/e069c37d3741de259d8bf745a4b9f2a3b35cdbd9

webservice/components/request_adapter.py
petrus-v: I'm going to test this new version on real environement

petrus-v:
I've squashed commits and rebase the PR, also I've move the way to change the state on the consumer record adding method on the abstract method to make it easier the way to overwrite it as discussed https://github.com/OCA/edi/pull/787#discussion_r1319477936

webservice/models/webservice_consumer_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+        compute="_compute_ws_response_content_filename"
+    )
+
+    def _compute_ws_response_content_filename(self):
+        for rec in self:
+            if rec.ws_response_date and rec.ws_response_status_code:
+                formatted_response_date = slugify(
+                    rec.ws_response_date.strftime(DATETIME_FORMAT)
+                )
+                rec.ws_response_content_filename = "response_%s_%s.json" % (
```
simahawk, line 41: Not sure assuming this is always json is ok. It can be pure txt or even html (unfortunately).
I'd leave it w/o ext for now. Not a blocker tho.

webservice/models/webservice_consumer_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    ws_response_content_filename = fields.Char(
+        compute="_compute_ws_response_content_filename"
+    )
+
+    def _compute_ws_response_content_filename(self):
+        for rec in self:
+            if rec.ws_response_date and rec.ws_response_status_code:
+                formatted_response_date = slugify(
+                    rec.ws_response_date.strftime(DATETIME_FORMAT)
```
simahawk, line 39:
```suggestion
                    fields.Datetime.to_string(rec.ws_response_date)
```

webservice/models/webservice_backend.py
```diff
@@ -71,8 +81,38 @@ def _valid_field_parameter(self, field, name):
...
+    def call(self, method, *args, consumer_record=None, **kwargs):
+        content = False
+        status_code = False
+        try:
+            content = getattr(self._get_adapter(), method)(*args, **kwargs)
+            status_code = 200
+        except HTTPError as request_error:
+            content = request_error.response.content
+            status_code = request_error.response.status_code
+            raise request_error from request_error
```
simahawk, line 108:
```suggestion
            content = request_error.response.content
            status_code = request_error.response.status_code
            raise
```

simahawk (commented):
@petrus-v overall LG.

PR #7, opened 2022-08-27
[14.0] webservice: improve call

simahawk:
'call' now accepts these explicit kw params:

* 'url': to override the request's url completely
* 'url_params': to pass interpolation params for url formatting.

PR #23, opened 2023-07-10
[16.0][MIG] webservice: Migration to 16.0

EvaSForgeFlow:
Standard migration


OriolMForgeFlow (approved):

OriolMForgeFlow (approved):
LGTM!

webservice/components/request_adapter.py
```diff
@@ -17,7 +17,7 @@ def _request(self, method, **kwargs):
             {"auth": self._get_auth(**kwargs), "headers": self._get_headers(**kwargs)}
         )
         request = requests.request(
-            method, self.collection.url.format(**kwargs), **new_kwargs
+            method, self.collection.url.format(**kwargs), **new_kwargs, timeout=None
```
LoisRForgeFlow: Include timeout in the new_kwargs update a couple of lines above, otherwise this will fail if timeout key is specified when calling _request method

simahawk (changes requested):
please fwd port changes from v14 too

OriolMForgeFlow:
Hi @simahawk , 
The FW Ports of #7 and #8 have been added. 
Also added the commits in v15 on #26 and #27

simahawk (approved):
LG. Thanks for your work :pray: 
Question: out of curiosity, for which purpose do you need this on v16? EDI or another use case?

simahawk:
Typo-ed module name before...

/ocabot migration webservice

OriolMForgeFlow:
> LG. Thanks for your work 🙏 Question: out of curiosity, for which purpose do you need this on v16? EDI or another use case?

Yes, we would like to use EDI in v16. We are also in the process of migrating it.

PR #31, opened 2024-01-02
[17.0][MIG] endpoint_route_handler: Migration to 17.0

nguyenminhchien:
Ref: BSRD-746


gurneyalex (approved):

simahawk (approved):
There are some things that could be cleaned up in the roadmap but no blocking points

PR #35, opened 2024-02-28
[16.0] [IMP] webservice: add support for oauth2

gurneyalex:


webservice/components/request_adapter.py
```diff
@@ -4,30 +4,51 @@
...
+
+
 class BaseRestRequestsAdapter(Component):
     _name = "base.requests"
     _webservice_protocol = "http"
     _inherit = "base.webservice.adapter"
 
     # TODO: url and url_params could come from work_ctx
     def _request(self, method, url=None, url_params=None, **kwargs):
+        Session = self._get_session_class()
```
simahawk: Wouldn't be easier to not modify this adapter, you should create a new one?
The protocol can be `http_oauth`

simahawk (changes requested):
We need tests

gurneyalex:
changing to draft: this implementation requests a new token for each api call, and some OAuth2 provider have an invoicing per token -> not acceptable.

webservice/components/request_adapter.py
gurneyalex: I rewrote the PR, I had come to the same conclusion. The result I feel is much cleaner.

webservice/README.rst
```diff
@@ -7,7 +7,7 @@ WebService
    !! This file is generated by oca-gen-addon-readme !!
    !! changes will be overwritten.                   !!
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-   !! source digest: sha256:303644ddb310b8c13e53c0cec4ea4f665b2ebb8a86b0c17bae78968c38633f4b
```
simahawk: NOTE: you can avoid these changes in PR by setting an env var: `SKIP=oca-gen-addon-readme`.
See https://github.com/OCA/web-api/blob/14.0/.github/workflows/pre-commit.yml#L38C17-L38C37

webservice/components/request_adapter.py
```diff
@@ -70,3 +75,106 @@ def _get_url(self, url=None, url_params=None, **kwargs):
...
+    _webservice_protocol = "http+oauth2"
+    _inherit = "base.requests"
+
+    def __init__(self, *args, **kw):
+        super().__init__(*args, **kw)
+        # cached value to avoid hitting the database each time we need the token
+        self._token = {}
+
+    def fetch_token(self):
+        pass
```
simahawk: a comment at least?

webservice/components/request_adapter.py
```diff
@@ -70,3 +75,106 @@ def _get_url(self, url=None, url_params=None, **kwargs):
...
+
+    def _is_token_valid(self, token):
+        """We consider that a token in valid if it has at least 10% of
+        its valid duration. So if a token has a validity of 1h, we will
+        renew it if we try to use it 6 minutes before its expiration date.
+        """
```
simahawk:
```suggestion
        """Validate given oauth2 token.

        We consider that a token in valid if it has at least 10% of
        its valid duration. So if a token has a validity of 1h, we will
        renew it if we try to use it 6 minutes before its expiration date.
        """
```

webservice/components/request_adapter.py
```diff
@@ -70,3 +75,106 @@ def _get_url(self, url=None, url_params=None, **kwargs):
...
+        """Return a valid oauth2 token.
+
+        The tokens are stored in the database, and we check if they are still
+        valid, and renew them if needed.
+        """
+        if self._is_token_valid(self._token):
+            return self._token
+        backend = self.collection
+        with backend.env.registry.cursor() as cr:
+            if backend.env.context.get("test_mode"):
```
simahawk: why this hack?

webservice/components/request_adapter.py
```diff
@@ -70,3 +75,106 @@ def _get_url(self, url=None, url_params=None, **kwargs):
...
+                    (backend.id,),
+                )
+                token_str = cr.fetchone()[0] or "{}"
+            token = json.loads(token_str)
+            if self._is_token_valid(token):
+                self._token = token
+                return token
+            else:
+                new_token = self._fetch_new_token(old_token=token)
+                if backend.env.context.get("test_mode"):
```
simahawk: same

webservice/models/webservice_backend.py
```diff
@@ -1,8 +1,10 @@
 # Copyright 2020 Creu Blanca
-# Copyright 2022 Camptocamp SA
+# Copyright 2022-2024 Camptocamp SA
```
simahawk: no need to update the year

simahawk (changes requested):

webservice/components/request_adapter.py
gurneyalex: Originally, to avoid a commit during the tests. I'll check if it is really needed.

webservice/components/request_adapter.py
gurneyalex: leftover -> deleting

webservice/components/request_adapter.py
gurneyalex: issue in the tests where the backend record is not flushed to disk and cannot be found by the SELECT, if I remember correctly. I'm removing this to see if I called `flush_model()` correctly at the time.

gurneyalex: https://github.com/OCA/web-api/actions/runs/8322619666/job/22770728397#step:8:204 is what I get.

SilvioC2C (approved):
LGTM overall

webservice/tests/test_oauth2.py
```diff
@@ -0,0 +1,101 @@
...
+                "access_token": "cool_token",
+                "expires_at": expires_timestamp,
+                "expires_in": duration,
+                "token_type": "Bearer",
+            },
+        )
+        responses.add(responses.GET, f"{self.url}endpoint", body="OK")
+
+        with mock_cursor(self.env.cr):
+            result = self.webservice.call("get", url=f"{self.url}endpoint")
```
simahawk, line 70: No need to test that commit has been called? Or that rollback is handled in case of errors?

gurneyalex: @simahawk  tests added in https://github.com/OCA/web-api/commit/ee889c409e005fd40160c97b541e57c83a13cba7

PR #36, opened 2024-02-28
[IMP] webservice: combine the url with collection's url

gurneyalex:
when a call is made with just a path, combine it with the collection's url

### DIFF
--- a/webservice/components/request_adapter.py
+++ b/webservice/components/request_adapter.py
@@ -65,8 +65,14 @@ class BaseRestRequestsAdapter(Component):
 
     def _get_url(self, url=None, url_params=None, **kwargs):
         if not url:
-            # TODO: if url is given, we should validate the domain
-            # to avoid abusing a webservice backend for different calls.
             url = self.collection.url
+        elif not url.startswith(self.collection.url):
+            if not url.startswith("http"):
+                url = f"{self.collection.url.rstrip('/')}/{url.lstrip('/')}"
+            else:
+                # TODO: if url is given, we should validate the domain
+                # to avoid abusing a webservice backend for different calls.
+                pass
+
         url_params = url_params or kwargs
         return url.format(**url_params)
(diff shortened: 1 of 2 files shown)


webservice/components/request_adapter.py
```diff
@@ -65,8 +65,14 @@ def _get_headers_for_api_key(self, **kw):
 
     def _get_url(self, url=None, url_params=None, **kwargs):
         if not url:
-            # TODO: if url is given, we should validate the domain
-            # to avoid abusing a webservice backend for different calls.
             url = self.collection.url
+        elif not url.startswith(self.collection.url):
+            if not url.startswith("http"):
+                url = self.collection.url + url
```
simahawk:
```suggestion
                url = f"{self.collection.url.rstrip('/')/{url.lstrip('/')"
```

simahawk: Question: why would you use this in favor of url params?

simahawk (changes requested):

webservice/components/request_adapter.py
gurneyalex: currently it is possible to use to send http requests on a different url (different name even) => I'm not fond of this but I don't want to break code that could use this "feature". 

In my case, I have a number of endpoints to access which are on "https://some.service.provider/api-prefix" and I find it nicer to use call() with only the part of the URL which is after this bit. That first part of the URL is going to be different in integration and production, but the value which is passed to call() is going to be stable. 

But maybe I'm missing soemthing obvious?

simahawk: but that's exactly what url_params are for...
The base url will be different by env: fine, you have the env conf for it.
You will have something like `https://sendto.me/prefix/{endpoint}/suffix` and you can pass `endpoint=foo` as arg.
That's the idea. However this is probably a nice improvement for when the endpoint is always at the end of the URL.

gurneyalex: @simahawk can you update your review?

simahawk (approved):

simahawk:
do we need an update on the readme?

gurneyalex:
> do we need an update on the readme?

I will make a PR enhancing the readme once #35 and #36 are merged.

PR #37, opened 2024-03-04
[14.0][FIX] endpoint_route_handler: Use dedicated cursor for registry

grindtildeath:
When the request cursor is used to instantiate the EndpointRegistry in the call to routing_map, the READ REPEATABLE isolation level will ensure that any value read from the DB afterwards, will be the same than when the first SELECT is executed.

This is breaking the oauth flow as the oauth token that is written at the beginning of the oauth process cannot be read by the cursor computing the session token, which will read an old value. Therefore when the session security check is performed, the session token is outdated as the new session token is computed using an up to date cursor.

By using a dedicated cursor to instantiate the EndpointRegistry, we ensure no read is performed on the database using the request cursor which will in turn use the updated value of the oauth token to compute the session token, and the security check will not fail.

### DIFF
--- a/endpoint_route_handler/models/ir_http.py
+++ b/endpoint_route_handler/models/ir_http.py
@@ -7,7 +7,7 @@ from itertools import chain
 
 import werkzeug
 
-from odoo import http, models
+from odoo import http, models, registry as registry_get
 
 from ..registry import EndpointRegistry
 
@@ -18,8 +18,8 @@ class IrHttp(models.AbstractModel):
     _inherit = "ir.http"
 
     @classmethod
-    def _endpoint_route_registry(cls, env):
-        return EndpointRegistry.registry_for(env.cr)
+    def _endpoint_route_registry(cls, cr):
+        return EndpointRegistry.registry_for(cr)
 
     @classmethod
     def _generate_routing_rules(cls, modules, converters):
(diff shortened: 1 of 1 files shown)


endpoint_route_handler/models/ir_http.py
```diff
@@ -41,20 +41,25 @@ def _endpoint_routing_rules(cls):
...
-            cls._routing_map = {}
-            cls._rewrite_len = {}
-            cls._endpoint_route_last_version = last_version
+        registry = registry_get(http.request.env.cr.dbname)
+        with registry.cursor() as cr:
+            last_version = cls._get_routing_map_last_version(cr)
+            if not hasattr(cls, "_routing_map"):
+                _logger.info(
+                    "routing map just initialized, store last update for this env"
+                )
```
simahawk: This will cause too much noise. Use `debug` if you want to keep it.
```suggestion
```

endpoint_route_handler/models/ir_http.py
```diff
@@ -32,7 +32,7 @@ def _generate_routing_rules(cls, modules, converters):
     @classmethod
     def _endpoint_routing_rules(cls):
         """Yield custom endpoint rules"""
-        e_registry = cls._endpoint_route_registry(http.request.env)
+        e_registry = cls._endpoint_route_registry(http.request.env.cr)
```
simahawk, line 35: shouldn't we use a new cr here too?

simahawk (changes requested):
Thanks for tracking down this issue :)

simahawk:
One last bit: can you add the explanation as a comment?

simahawk:
I'll apply the changes

endpoint_route_handler/models/ir_http.py
grindtildeath: It doesn't seem to be required for Oauth flow, so I don't know :shrug:

simahawk (approved):
Live on prod since 4 months.

PR #42, opened 2024-04-28
[16.0][ADD] webservice_server_env: make server_environment an optional dependency

dreispt:
- [ADD] webservice_server_env: make server_environment an optional dependency
- [REF] webservice: make server_environment an optional dependency

PR #47, opened 2024-06-17
[FIX] webservice: WARNING message in logs

gurneyalex:
The use of a compute method on `oauth2_flow` when this field is touched by the server environment mixin causes it to be defined twice as computed, with differents settings, and this ultimately causes a warning message in the logs:

```
WARNING odoo odoo.modules.registry: webservice.backend: inconsistent 'compute_sudo' for computed fields: protocol, url, auth_type, username, password, api_key, api_key_header, oauth2_flow, oauth2_clientid, oauth2_client_secret, oauth2_token_url, oauth2_authorization_url, oauth2_audience, oauth2_scope, content_type
```

We fix this by using an old fashioned onchange declaration on the `auth_type` field.

### DIFF
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -123,8 +123,8 @@ class WebserviceBackend(models.Model):
 
     @api.onchange("auth_type")
     def _onchange_oauth2_auth_type(self):
-        # reset the auth2_flow when auth_type is not oaut2
-        # using a compute method interfers with the server environment mixin
+        # reset the oauth2_flow when auth_type is not oauth2
+        # using a compute method interferes with the server environment mixin
         for rec in self:
             if rec.auth_type != "oauth2":
                 rec.oauth2_flow = False
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -40,7 +40,6 @@ class WebserviceBackend(models.Model):
         ],
         readonly=False,
         store=True,
-        compute="_compute_oauth2_flow",
     )
     oauth2_clientid = fields.Char(string="Client ID", auth_type="oauth2")
     oauth2_client_secret = fields.Char(string="Client Secret", auth_type="oauth2")
@@ -122,8 +121,10 @@ class WebserviceBackend(models.Model):
             protocol += f"+{self.auth_type}-{self.oauth2_flow}"
         return protocol
 
-    @api.depends("auth_type")
-    def _compute_oauth2_flow(self):
+    @api.onchange("auth_type")
+    def _onchange_oauth2_auth_type(self):
+        # reset the auth2_flow when auth_type is not oaut2
+        # using a compute method interfers with the server environment mixin
         for rec in self:
             if rec.auth_type != "oauth2":
                 rec.oauth2_flow = False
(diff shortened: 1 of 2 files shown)


webservice/models/webservice_backend.py
```diff
@@ -40,7 +40,6 @@ class WebserviceBackend(models.Model):
         ],
         readonly=False,
         store=True,
```
StephaneMangin: Could be removed also i presume

StephaneMangin (approved):

webservice/models/webservice_backend.py
```diff
@@ -122,8 +121,10 @@ def _get_adapter_protocol(self):
             protocol += f"+{self.auth_type}-{self.oauth2_flow}"
         return protocol
 
-    @api.depends("auth_type")
-    def _compute_oauth2_flow(self):
+    @api.onchange("auth_type")
+    def _onchange_oauth2_auth_type(self):
```
grindtildeath: IMO it's a breaking change going from a compute function (that will be triggered anytime auth_type is changed) to an onchange function (that will be triggered only on the UI).

Can't you instead redefine the `compute` slot of the field in the server env module to `False`?

gurneyalex: I still want this to be managed by server env.

webservice/models/webservice_backend.py
```diff
@@ -122,8 +121,10 @@ def _get_adapter_protocol(self):
...
+        # reset the auth2_flow when auth_type is not oaut2
+        # using a compute method interfers with the server environment mixin
```
henrybackman:
```suggestion
        # reset the oauth2_flow when auth_type is not oauth2
        # using a compute method interferes with the server environment mixin
```

webservice/models/webservice_backend.py
simahawk: I tend to agree

grindtildeath: I got confused, the model is inheriting from server env in this module, so you can ignore my comment

simahawk: I still think that it would be better to take full control instead.
Can't we control it by improving or overriding `_server_env_transform_field_to_read_from_env` ?

gurneyalex: To be checked too: https://github.com/OCA/web-api/pull/42

webservice/models/webservice_backend.py
gurneyalex: hmmm maybe yes, depending on how things turn out.

SilvioC2C:
@gurneyalex I've opened a new PR to speed up the fix: https://github.com/OCA/web-api/pull/63

@grindtildeath @simahawk @henrybackman @StephaneMangin would you mind reviewing the new PR?

webservice/models/webservice_backend.py
```diff
@@ -177,3 +169,10 @@ def _server_env_fields(self):
         }
         webservice_fields.update(base_fields)
         return webservice_fields
+
+    def _compute_server_env(self):
+        # OVERRIDE: reset ``oauth2_flow`` when ``auth_type`` is not "oauth2", even if
+        # defined otherwise in server env vars
+        res = super(WebserviceBackend, self)._compute_server_env()
```
rvalyi, line 176: I think it should be just super()

simahawk:
https://github.com/OCA/web-api/pull/63

PR #48, opened 2024-06-21
[16.0][IMP] webservice: allow to return response object on http calls

JordiMForgeFlow:
you may need to use information on the headers of the response

### DIFF
--- a/webservice/components/request_adapter.py
+++ b/webservice/components/request_adapter.py
@@ -24,6 +24,7 @@ class BaseRestRequestsAdapter(Component):
     # TODO: url and url_params could come from work_ctx
     def _request(self, method, url=None, url_params=None, **kwargs):
         url = self._get_url(url=url, url_params=url_params)
+        content_only = kwargs.pop("content_only", True)
         new_kwargs = kwargs.copy()
         new_kwargs.update(
             {
@@ -35,7 +36,9 @@ class BaseRestRequestsAdapter(Component):
         # pylint: disable=E8106
         request = requests.request(method, url, **new_kwargs)
         request.raise_for_status()
-        return request.content
+        if content_only:
+            return request.content
+        return request
 
     def get(self, **kwargs):
         return self._request("get", **kwargs)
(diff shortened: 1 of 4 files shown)


MiquelRForgeFlow (approved):

HviorForgeFlow (approved):
Code review :+1:

simahawk (commented):
Do we really need a field for this?
Can't we have a param to achieve the same?
Maybe you can explain a bit more about your use case?

JordiMForgeFlow:
@simahawk you mean to pass it as a parameter in the kwargs of the `_request` method?

The use case is basically that I need to extract some information of the headers of the response, so I need the result of the `_request` to not be only the content

simahawk:
> @simahawk you mean to pass it as a parameter in the kwargs of the `_request` method?

Yes, eg: `content_only=True` or `full_response=False`.

> The use case is basically that I need to extract some information of the headers of the response, so I need the result of the `_request` to not be only the content

But it means that you don't do this programmatically and you are calling `backend.call` directly which means you have full control over the call, right?

JordiMForgeFlow:
@simahawk yes, I am calling backend.call directly, so I should be able to pass it as a parameter. I will propose the change.

HviorForgeFlow (approved):
Some reference about content_only kwarg in the README would be nice :)

webservice/components/request_adapter.py
```diff
@@ -24,6 +24,7 @@ class BaseRestRequestsAdapter(Component):
     # TODO: url and url_params could come from work_ctx
     def _request(self, method, url=None, url_params=None, **kwargs):
         url = self._get_url(url=url, url_params=url_params)
+        content_only = kwargs.pop("content_only", True)
```
MiquelRForgeFlow, line 27: reference this in readme :)

JordiMForgeFlow:
@simahawk could we get this one merged? :)

JordiMForgeFlow:
@simahawk could we get this one merged?

etobella (approved):

PR #56, opened 2024-07-23
[14.0] endpoint: add basic mixin view

simahawk:
As the mixin can be inherited by many other models w/ their view an isolated basic view is required to avoid pollution and breakage of inherited views when extending the views of endpoint.endpoint.

In particular, this solves a conflict that came w/ https://github.com/OCA/web-api-contrib/pull/1

Companion PR for EDI: https://github.com/OCA/edi/pull/1012

### DIFF
--- a/endpoint_cache/views/endpoint_view.xml
+++ b/endpoint_cache/views/endpoint_view.xml
@@ -3,9 +3,9 @@
      License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
 <odoo>
 
-    <record model="ir.ui.view" id="endpoint_endpoint_form_view">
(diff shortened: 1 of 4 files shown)


endpoint_cache/views/endpoint_view.xml
```diff
@@ -5,7 +5,7 @@
 
     <record model="ir.ui.view" id="endpoint_endpoint_form_view">
```
henrybackman:
```suggestion
    <record model="ir.ui.view" id="endpoint_mixin_form_view">
```

henrybackman (approved):

endpoint_cache/views/endpoint_view.xml
simahawk: applied :)

PR #61, opened 2024-08-21
[17.0][FIX] endpoint_route_handler: acl warning

qgroulard:
Empty group_id in acl is deprecated.
see https://github.com/odoo/odoo/commit/ce1a4a2ee00fadf3c09b9dc77baa74c83375331b


oerp-odoo:
@qgroulard according to @simahawk we should no longer need "public" access for that model. So I guess we should simply get rid of those ACLs?

related: https://github.com/OCA/web-api/issues/62

qgroulard:
> @qgroulard according to @simahawk we should no longer need "public" access for that model. So I guess we should simply get rid of those ACLs?
> 
> related: #62

Fine by me, but I'm not really comfortable with that. I prefer to let @simahawk remove the tool model and the according ACLs.
Until the tool model is removed, this PR is still useful.

simahawk (approved):

PR #63, opened 2024-09-06
[FIX] webservice: WARNING message in logs

SilvioC2C:
The use of a compute method on ``oauth2_flow`` when this field is touched by the server environment mixin causes it to be defined twice as computed, with differents settings, and this ultimately causes a warning message in the logs:

```
WARNING odoo odoo.modules.registry: webservice.backend: inconsistent 'compute_sudo' for computed fields: protocol, url, auth_type, username, password, api_key, api_key_header, oauth2_flow, oauth2_clientid, oauth2_client_secret, oauth2_token_url, oauth2_authorization_url, oauth2_audience, oauth2_scope, content_type
```

We fix this by overriding method ``compute_server_env()``

Supersedes https://github.com/OCA/web-api/pull/47

### DIFF
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -39,8 +39,6 @@ class WebserviceBackend(models.Model):
             ("web_application", "Web Application (Authorization Code Grant)"),
         ],
         readonly=False,
-        store=True,
-        compute="_compute_oauth2_flow",
     )
     oauth2_clientid = fields.Char(string="Client ID", auth_type="oauth2")
     oauth2_client_secret = fields.Char(string="Client Secret", auth_type="oauth2")
@@ -122,12 +120,6 @@ class WebserviceBackend(models.Model):
             protocol += f"+{self.auth_type}-{self.oauth2_flow}"
         return protocol
 
-    @api.depends("auth_type")
-    def _compute_oauth2_flow(self):
-        for rec in self:
-            if rec.auth_type != "oauth2":
-                rec.oauth2_flow = False
-
     @api.depends("auth_type", "oauth2_flow")
     def _compute_redirect_url(self):
         get_param = self.env["ir.config_parameter"].sudo().get_param
@@ -177,3 +169,10 @@ class WebserviceBackend(models.Model):
         }
         webservice_fields.update(base_fields)
         return webservice_fields
+
+    def _compute_server_env(self):
+        # OVERRIDE: reset ``oauth2_flow`` when ``auth_type`` is not "oauth2", even if
+        # defined otherwise in server env vars
+        res = super()._compute_server_env()
+        self.filtered(lambda r: r.auth_type != "oauth2").oauth2_flow = None
+        return res
--- a/webservice/tests/test_oauth2.py
+++ b/webservice/tests/test_oauth2.py
@@ -4,10 +4,16 @@
 import json
 import os
 import time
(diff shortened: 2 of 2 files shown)


StephaneMangin (approved):

webservice/models/webservice_backend.py
```diff
@@ -177,3 +169,10 @@ def _server_env_fields(self):
         }
         webservice_fields.update(base_fields)
         return webservice_fields
+
+    def _compute_server_env(self):
+        # OVERRIDE: reset ``oauth2_flow`` when ``auth_type`` is not "oauth2", even if
+        # defined otherwise in server env vars
```
simahawk, line 175: Can you provide the full description of the issue here?

Q: as there's no dependency on `auth_type` here (of course), is this playing nice within the form too? Meaning: if you don't have an env conf, do you get a proper reset of the value when changing the `auth_type`?

webservice/tests/test_oauth2.py
```diff
@@ -213,3 +216,33 @@ def test_fetch_token_from_auth(self):
         adapter = self.webservice._get_adapter()
         token = adapter._fetch_token_from_authorization(code)
         self.assertEqual("cool_token", token["access_token"])
+
+    def test_oauth2_flow_compute(self):
+        # Test with current configuration
+        env = os.environ
```
simahawk: Please mock the env properly in this way https://github.com/OCA/queue/blob/16.0/test_queue_job/tests/test_delay_mocks.py#L345

webservice/tests/test_oauth2.py
```diff
@@ -213,3 +216,33 @@ def test_fetch_token_from_auth(self):
...
+        server_env_mixin.serv_config = server_env._load_config()  # Reload env vars
+        self.webservice.invalidate_recordset()  # Force recomputation when reading
+        self.assertEqual(self.webservice.oauth2_flow, "web_application")
+
+        # Update configuration: ``auth_type`` is changed to ``none``
+        env["SERVER_ENV_CONFIG"] = env["SERVER_ENV_CONFIG"].replace(
+            "auth_type = oauth2", "auth_type = none"
+        )
+        server_env_mixin.serv_config = server_env._load_config()  # Reload env vars
+        self.webservice.invalidate_recordset()  # Clear cache => read forces compute
```
simahawk: This tells me that the answer to the question above is "no, we lose the update on the form".

simahawk (changes requested):
I would reconsider adding back the onchange as for #47 now that you have the auto-compute properly done.

webservice/models/webservice_backend.py
SilvioC2C: > is this playing nice within the form too?

Yep, tested in UI.

webservice/tests/test_oauth2.py
SilvioC2C: No, we don't lose the update on the form: it is lost in tests because Odoo will fetch the cached value, but that doesn't happen when reading the value in UI.

SilvioC2C:
@simahawk another review, please? :smile:

simahawk (approved):
LG

simahawk:
@SilvioC2C pls squash

SilvioC2C:
@simahawk can we merge this?

gurneyalex (approved):

PR #65, opened 2024-10-21
[18.0][mig] endpoint_route_handler

samirGuesmi:
Bonjour @simahawk, I would like to get your input here. 

in Odoo 18: Adds a readonly parameter, allowing routes to use a read-only replica of the database when appropriate. This is beneficial for endpoints that only perform read operations, such as querying records, and it helps to improve performance by reducing the load on the primary database. While the readonly parameter is intended to be optional, test cases encountered failures due to a missing readonly key, as seen in the error:

site-packages/odoo/http.py", line 1888, in _serve_db
    readonly = rule.endpoint.routing['readonly']
               ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: 'readonly'


simahawk:
@samirGuesmi thanks for your work and for the info. I'll have a look in the following days, merci :)

simahawk:
The flag and failing tests have been handled here https://github.com/OCA/web-api/pull/69
You can still help w/ a review, if you want! Thanks!

PR #66, opened 2024-10-23
[18.0][MIG] webservice

JordiMForgeFlow:
- Standard V18 Migration
- Allow empty content type in webservice backend: when working with python requests library to POST a multipart/form-data type, we should not explicitly set the content-type in the headers as the library will take care of putting the correct content-type and also the boundaries. Otherwise it will not do it. (https://stackoverflow.com/questions/12385179/how-to-send-a-multipart-form-data-with-requests-in-python).
- Do not log warning on tests: the warning otherwise causes the CI to fail due to the warnings.
- FW of https://github.com/OCA/web-api/pull/48


JordiMForgeFlow:
CC @etobella @simahawk

GuillemCForgeFlow (approved):
code review 👍🏿

MiquelRForgeFlow (approved):

etobella (approved):
/ocabot merge nobump

PR #69, opened 2024-11-08
[18.0][MIG] endpoint_route_handler: Migration to 18.0

cormaza:
Supersed https://github.com/OCA/web-api/pull/65

Added parameter readonly in route definition, by default false to keep standard behavior


cormaza:
Thanks for this module @simahawk, please let me know if everything is fine

simahawk:
Thanks for your work!
LG but I think we miss this PR that is not fwd ported yet to v17 https://github.com/OCA/web-api/pull/52
Would you mind porting it to 17 and then cherry-pick the commit here?

cormaza:
> Thanks for your work! LG but I think we miss this PR that is not fwd ported yet to v17 #52 Would you mind porting it to 17 and then cherry-pick the commit here?

I'm no really sure, because in version 16 you have a class method, then in 17 version is a regular function, maybe we can include after forward ported  to 17 version I can include in this MR, wdyt?

simahawk:
Sure, we have to port all changes to v17 first. You can try https://github.com/OCA/oca-port :wink:

simahawk:
And there's a fix for it that just landed in 16 https://github.com/OCA/web-api/pull/77

lmignon (approved):

simahawk (approved):
I'd like to get rid of the `.tool` as I've found an alternative but I don't have time to provide specs now... and this is blocking other modules.
Thanks!

simahawk:
FYI for existing records you must force sync otherwise the new param `readonly` won't be stored and the endpoint will be broken.
Here are specific fixes for depending modules:

https://github.com/OCA/web-api/pull/105
https://github.com/OCA/edi-framework/pull/184

Manual work around:
1. archive
2. sync
3. unarchive
4. sync

@lmignon maybe you need that for existing fastapi.endpoint too

PR #70, opened 2024-11-19
[16.0][FIX]  authorization

baringji:
Add scope param in the authorization url when using oauth2.
Fix 'webservice.backend' object has no attribute get_access_action.

### DIFF
--- a/webservice/components/request_adapter.py
+++ b/webservice/components/request_adapter.py
@@ -240,6 +240,7 @@ class WebApplicationOAuth2RestRequestsAdapter(Component):
         with OAuth2Session(
             client=client,
             redirect_uri=oauth_params.get("redirect_url"),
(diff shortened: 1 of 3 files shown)


simahawk:
@gurneyalex can you pls review this?
@baringji thanks for your contrib!
Could you please have a look at our [commit guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message).
Namely:
1. drop the odoo version
2. add module name
3. shorten the msg to be more synthetic and relevant and keep further info for the commit description (eg: `[FIX] webservice: webservice.backend.get_access_token AttributError`

Also, tests need an update very likely :)

santostelmo:
@gurneyalex I think this fix make sense. The scope should be passed as parameter

PR #74, opened 2024-11-26
[17.0][IMP] webservice: use refresh token

SilvioC2C:


simahawk:
It starts to be very crowded... This makes me think that we should move out this feature to its own module.

PR #77, opened 2024-12-09
[FIX] endpoint_route_handler: Rollback dedicated routing cursor

paradoxxxzero:
Unless I am missing something, we should not commit this transaction (`registry.cursor()` commits at [`__exit__ `](https://github.com/odoo/odoo/blob/16.0/odoo/sql_db.py#L176)).

This fix the failing tests in rest-framework: https://github.com/OCA/rest-framework/issues/470

### DIFF
--- a/endpoint_route_handler/models/ir_http.py
+++ b/endpoint_route_handler/models/ir_http.py
@@ -58,7 +58,8 @@ class IrHttp(models.AbstractModel):
         # which will in turn use the updated value of the oauth token to compute
         # the session token, and the security check will not fail.
         registry = registry_get(http.request.env.cr.dbname)
-        with registry.cursor() as cr:
+        cr = registry.cursor()
+        try:
             last_version = cls._get_routing_map_last_version(cr)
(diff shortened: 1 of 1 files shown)


simahawk (approved):
LG, good catch :)

endpoint_route_handler/models/ir_http.py
```diff
@@ -71,6 +72,10 @@ def routing_map(cls, key=None):
                 cls._routing_map = {}
                 cls._rewrite_len = {}
                 cls._endpoint_route_last_version = last_version
+        finally:
+            cr.rollback()
```
simahawk, line 79: can you add a comment here pls?
Mention that the new cr is only to bypass transaction isolation during a read  or smth similar.

paradoxxxzero: Done

JordiMForgeFlow (approved):

lmignon (approved):
Thank you for the fix @paradoxxxzero LGTM

PR #80, opened 2024-12-23
[14.0][IMP] endpoint: support usage of querystring parameters

dreispt:


dreispt:
Review is appreciated but please don't merge yet, there might be a few changes.

dreispt:
Ready for review.

endpoint/models/endpoint_mixin.py
```diff
@@ -76,8 +76,10 @@ def _default_code_snippet_docs(self):
         * Response
         * werkzeug
         * exceptions
+        * params
```
simahawk, line 79:
```suggestion
        * querystring_params
```

endpoint/models/endpoint_mixin.py
```diff
@@ -134,10 +136,11 @@ def _code_snippet_log_func(self, message, level="info"):
                 ),
             )
 
-    def _handle_exec__code(self, request):
+    def _handle_exec__code(self, request, params=None):
```
simahawk, line 139:
```suggestion
    def _handle_exec__code(self, request, querystring_params=None):
```

endpoint/controllers/main.py
```diff
@@ -17,7 +17,7 @@ def _handle_endpoint(self, env, model, endpoint_route, **params):
         if not endpoint:
             raise NotFound()
         endpoint._validate_request(request)
-        result = endpoint._handle_request(request)
+        result = endpoint._handle_request(request, params=params)
```
simahawk, line 20:
```suggestion
        result = endpoint._handle_request(request, querystring_params=params)
```

endpoint/models/endpoint_mixin.py
```diff
@@ -134,10 +136,11 @@ def _code_snippet_log_func(self, message, level="info"):
...
         if not self._code_snippet_valued():
             return {}
         eval_ctx = self._get_code_snippet_eval_context(request)
+        eval_ctx["params"] = params or {}
```
simahawk, line 143:
```suggestion
        eval_ctx["querystring_params"] = querystring_params or {}
```

endpoint/models/endpoint_mixin.py
```diff
@@ -184,14 +187,18 @@ def _get_handler(self):
                 _("Missing handler for exec mode %s") % self.exec_mode
             )
 
-    def _handle_request(self, request):
+    def _handle_request(self, request, params=None):
```
simahawk, line 190:
```suggestion
    def _handle_request(self, request, querystring_params=None):
```

endpoint/models/endpoint_mixin.py
```diff
@@ -184,14 +187,18 @@ def _get_handler(self):
...
         # Switch user for the whole process
         self_with_user = self
         if self.exec_as_user_id:
             self_with_user = self.with_user(user=self.exec_as_user_id)
         handler = self_with_user._get_handler()
         try:
-            res = handler(request)
+            # In case the handler does not support params
```
simahawk, line 197: I doubt we have any other handler ATM but just in case... ok, let's keep this but I would add a separated commit for backward compat that we can easily drop on migration. 
Something like:

```
[imp] endpoint: support querystring params
[imp] endpoint: backward compat for querystring params

Old handlers might not accept this new kw args
Trash this commit in future versions. Eg: v18.
```

endpoint/models/endpoint_mixin.py
```diff
@@ -184,14 +187,18 @@ def _get_handler(self):
...
+            if params:
+                res = handler(request, params=params)
```
simahawk, line 199:
```suggestion
            if querystring_params:
                res = handler(request, querystring_params=querystring_params)
```

simahawk (changes requested):
Thank you! Something I was thinking about since a while ;)

Just 1 remark about the naming: I prefer to make explicit everywhere down the stack, as `params` is quite an abused term :wink: 

Also, would be nice to see a test.

PR #91, opened 2025-02-19
[14.0] Add endpoint_cache_preheat

simahawk:
Depends  on
- [x] https://github.com/OCA/web-api/pull/90

### DIFF
--- a/endpoint_cache_preheat/models/endpoint_mixin.py
+++ b/endpoint_cache_preheat/models/endpoint_mixin.py
@@ -1,0 +1,56 @@
+# Copyright 2025 Camptocamp SA
+# @author: Simone Orsi <simone.orsi@camptocamp.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import logging
+
+from odoo import api, fields, models
+
+_logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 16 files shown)


endpoint_cache_preheat/models/endpoint_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+
+    _inherit = "endpoint.mixin"
+
+    cache_preheat = fields.Boolean(
+        string="Automatic cache pre-heat",
+        help="If checked, cache will be pre-heated by a cron "
+        "according to the selected policy. "
+        "Cache generation will be done in a queue job.",
+    )
+    cache_preheat_ts = fields.Datetime(readonly=True)
```
SilvioC2C:
```suggestion
    cache_preheat_timestamp = fields.Datetime(readonly=True)
```

SilvioC2C (approved):

endpoint_cache_preheat/models/endpoint_mixin.py
simahawk: I've improved the label

SilvioC2C: Yeah, it's the field name itself that could be more readable :smile:

PR #93, opened 2025-03-05
[16.0][IMP] endpoint: add hmac and hashlib support

ivs-cetmix:
Why?
----
We would like to use this module to get webhooks from GitHub. However GitHub uses hashed passwords, which requires additional libraries.

Solution
--------

Make `hashlib` and `hmac` available in the eval context.

### DIFF
--- a/endpoint/models/endpoint_mixin.py
+++ b/endpoint/models/endpoint_mixin.py
@@ -12,6 +12,31 @@ from odoo.tools import safe_eval
 
 from odoo.addons.rpc_helper.decorator import disable_rpc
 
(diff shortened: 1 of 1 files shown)


GabbasovDinar (approved):

Aldeigja (approved):
Functionally LGTM, we're already using it in production.

ivs-cetmix:
Ciao @simahawk, would appreciate your review and merge, we are going to update 17.0 as well. Grazie!

simahawk:
LG

/ocabot merge minor

PR #98, opened 2025-04-24
[18.0][ADD] endpoint_auth_bearer

xtanuiha:
Provide API key authorization in bearer mode for endpoints.

Ref: https://github.com/odoo/odoo/pull/180386


xtanuiha:
@pedrobaeza @simahawk 
I've added a new module called `endpoint_auth_bearer`. Please help review the code when you have time. Thanks!

PR #101, opened 2025-06-15
[18.0][ADD] webservice_server_env: make server_environment an optional dependency

dreispt:
Forward port of #42

- [REF] webservice: server_environment should not be a required dependency
- [ADD] webservice_server_env: make server_environment an optional dependency

PR #102, opened 2025-06-15
[18.0][ADD] webservice_base: lighter version of webservice module

dreispt:
Refactoring of `webservice` module to remove dependency on `component`.
Depends on (and includes) #101

New commits:
- [ADD] webservice_base


webservice_base/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,10 @@
+This module creates WebService frameworks to be used globally.
+
+It introduces support for HTTP Request protocol.
+The webservice HTTP call returns by default the content of the response.
+A context 'content_only' can be passed to get the full response object.
+
+It comes with no additional dependencies.
+To rely also on the `components` objetcs, see the `webservid` addon odule instead.
```
em230418: "addon module instead"

em230418: and "objects", not "objetcs"

ivs-cetmix:
Hi @dreispt , is there any functional documentation available for this module?

dreispt:
@ivs-cetmix see #119

flotho:
Hi, 
Thanks for this PR,
Any chance to rebase to trigger CI/CD ?

webservice_base/models/webservice_request_adapter.py
```diff
@@ -0,0 +1,79 @@
...
+
+from odoo import models
+
+from ..utils import sanitize_url_for_log
+
+_logger = logging.getLogger(__name__)
+
+
+class WebserviceRequestAdapter(models.AbstractModel):
+    _name = "webservice.request.adapter"
```
simahawk, line 18: TBH I don't like this implementation.
It does not support different adapters for different use cases.
It should be possible to register specific adapters for the protocol.
Plus, the methods here should be renamed to not clash w/ core methods (existing or new to come).

simahawk (changes requested):

webservice_base/models/webservice_request_adapter.py
dreispt: There are two approaches used in core Odoo to support adapter/plugin like features:

1. Adapter specific logic guarded by an "if" statement in extension methods.
2. Base method routes to adapter specific logic using dynamic calls to methods.

The 1. approach has some limitations; the 2. approach is the *de facto* standard, as seen in delivery carriers and , payment providers.

I can work on expanding this idea  to include the routing mechanism.

simahawk:
FTR I'm working on an alternative solution.

PR #103, opened 2025-06-15
[18.0][ADD] webservice_base_oauth

dreispt:
Depends on (and includes) #102 

New commits:
- [ADD] webservice_base_oauth


dreispt:
The token logic need to be refactored to use Odoo native patterns.
I'm not sure I'll take this on if there isn't a reasonable chance for any of it to be merged...

PR #107, opened 2025-07-16
[18.0] endpoint_route_handler: wipe cache only once on sync

simahawk:
Only for v18.


TDu (approved):

simahawk:
Tested in the ctx of shopfloor :+1: 
/ocabot merge minor

PR #108, opened 2025-07-16
[18.0] endpoint: fix migration to v18

simahawk:
A new route parameter has been introduced and for existing endpoints to work OOTB we must resync.


simahawk:
Tested manually on real project.
/ocabot merge patch

PR #114, opened 2025-09-22
[17.0][FWD] `endpoint`: forward-port querystring params usage

SilvioC2C:
Forward-port of https://github.com/OCA/web-api/pull/80

Also includes renaming usages of `params` to `querystring_params`

### DIFF
--- a/endpoint/models/endpoint_mixin.py
+++ b/endpoint/models/endpoint_mixin.py
@@ -105,7 +105,7 @@ class EndpointMixin(models.AbstractModel):
             'sha3_512', 'shake_128', 'shake_256', 'blake2b',
             'blake2s', 'md5', 'new'
         * hmac: Python 'hmac' library. Use 'new' to create HMAC objects.
-        * params
+        * querystring_params
 
         Must assign a ``result`` variable, with either an instance of ``Response``,
         or a dict containiny any of the keys:
@@ -189,11 +189,11 @@ class EndpointMixin(models.AbstractModel):
                 ),
             )
 
-    def _handle_exec__code(self, request, params=None):
+    def _handle_exec__code(self, request, querystring_params=None):
         if not self._code_snippet_valued():
             return {}
         eval_ctx = self._get_code_snippet_eval_context(request)
-        eval_ctx["params"] = params or {}
+        eval_ctx["querystring_params"] = querystring_params or {}
         snippet = self.code_snippet
         safe_eval.safe_eval(snippet, eval_ctx, mode="exec", nocopy=True)
         result = eval_ctx.get("result")
@@ -241,7 +241,7 @@ class EndpointMixin(models.AbstractModel):
                 _("Missing handler for exec mode %s") % self.exec_mode
             ) from e
 
-    def _handle_request(self, request, params=None):
+    def _handle_request(self, request, querystring_params=None):
         # Switch user for the whole process
         self_with_user = self
         if self.exec_as_user_id:
@@ -249,8 +249,8 @@ class EndpointMixin(models.AbstractModel):
         handler = self_with_user._get_handler()
         try:
             # In case the handler does not support params
-            if params:
-                res = handler(request, params=params)
+            if querystring_params:
+                res = handler(request, querystring_params=querystring_params)
             else:
                 res = handler(request)
         except self._bad_request_exceptions() as orig_exec:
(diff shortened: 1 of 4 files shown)


SilvioC2C:
NB: tests are failing for reasons unrelated to these commits

ivantodorovich (approved):

jcoux (approved):

dreispt (approved):

endpoint/models/endpoint_mixin.py
```diff
@@ -238,14 +241,18 @@ def _get_handler(self):
...
-    def _handle_request(self, request):
+    def _handle_request(self, request, querystring_params=None):
         # Switch user for the whole process
         self_with_user = self
         if self.exec_as_user_id:
             self_with_user = self.with_user(user=self.exec_as_user_id)
         handler = self_with_user._get_handler()
         try:
-            res = handler(request)
+            # In case the handler does not support params
```
simahawk, line 251: I would simply let it fail. Or if you really want you can catch a `TypeError` and log it. But I would just let it break. Beside, I doubt anyone did ever implement a different handler :)

simahawk (changes requested):
can we have a test update?

SilvioC2C:
@simahawk would it be better to make a commit in the v14 PR and port it here, or would suggest a commit here for simplicity?

simahawk:
we can backport it later

SilvioC2C:
@simahawk mind re-opening/re-reviewing this PR?

endpoint/models/endpoint_mixin.py
simahawk: to be answered

endpoint/tests/test_endpoint_controller.py
```diff
@@ -77,3 +77,7 @@ def test_call6(self):
     def test_call7(self):
         response = self.url_open("/demo/bad_method", data="ok")
         self.assertEqual(response.status_code, 405)
+
+    def test_call8(self):
```
simahawk, line 81: I'm not sure to understand how this test is supposed to work. You should pass params in the query string no?

endpoint/models/endpoint_mixin.py
```diff
@@ -187,10 +189,11 @@ def _code_snippet_log_func(self, message, level="info"):
                 ),
             )
 
-    def _handle_exec__code(self, request):
+    def _handle_exec__code(self, request, querystring_params=None):
         if not self._code_snippet_valued():
             return {}
         eval_ctx = self._get_code_snippet_eval_context(request)
+        eval_ctx["querystring_params"] = querystring_params or {}
```
simahawk, line 196: do we really need this? We have the request and as such we have `request.params` that contains the qs params.

simahawk (changes requested):

yankinmax:
Hello @SilvioC2C 
What is the status here?
Can you pls address reviewer's suggestion/questions?
Thank you in advance

endpoint/tests/test_endpoint_controller.py
yankinmax: The test only used the URL path converters (<string:param1>/<int:param2>), so it never actually checked a real query string.
I will update it to test `/demo/test/8?extra=1` and assert extra shows up in the merged `querystring_params` alongside the path `params`.

yankinmax:
Hello @simahawk from what I see it's a forward-port and the questions you've asked should be actually covered by the:
- [ ]  https://github.com/OCA/web-api/pull/80

I've fixed the test you requested.
Can you pls take a look and possibly unlock this PR to get meged?

simahawk:
@yankinmax TBH I even lost the WHY this is necessary.

Doesn't solve the one genuine gap. If you actually want querystring-only values (excluding path converters/form), neither `request.params` nor this PR's `querystring_params` give you that — both collapse everything into one dict. That distinction only exists in `request.httprequest.args` (a werkzeug MultiDict), which also happens to be the only way to get repeated keys via `.getlist()` — something neither `request.params` nor the PR's dict supports.

IMO we should only document how to do it.

In fact, there's already a test that shows that we can get params from the request w/ this demo record (from v18)

```
    <record id="endpoint_demo_6" model="endpoint.endpoint">
        <field name="name">Demo Endpoint 6</field>
        <field name="route">/demo/value_from_request</field>
        <field name="request_method">GET</field>
        <field name="auth_type">public</field>
        <field name="exec_as_user_id" ref="base.user_demo" />
        <field name="exec_mode">code</field>
        <field name="code_snippet">
result = {"response": Response(request.params.get("your_name", ""))}
        </field>
    </record>
```

@SilvioC2C @dreispt as you worked on the initial implementation, can you tell me what cannot be achieved w/ the current state?

SilvioC2C:
@simahawk TBH I didn't dig into this too much, I merely forward-ported #80 (including the renaming you suggested there) :smile: 

I'll leave the mic to @dreispt

simahawk:
closing for now. Feel free to reopen if there's any good reason to keep it ;)

PR #118, opened 2025-11-30
[19.0][endpoint-route-handler] - Migrated endpoint route hadnler to be used with fastapi 

chroma-fgobea:
Had to comment a function that avoided being able to get to the instance, Dont know if that function is needed for something else, but fastapi module is working this way with no issues.

If that function has an objective feel free to tell me which was and how to test it and i will be eager to finish the job and migrate this module to 19.0


chroma-fgobea:
Just saw this was a dup of another PR. 

Probably he did a better job!

PR #121, opened 2025-12-21
[18.0][IMP] Add Bearer token authentication support to endpoint

xtanuiha:
- Add Bearer token authentication support to endpoint
- Improve endpoint views.


simahawk (changes requested):
Thanks for you contrib and sorry for the late review.

Can you please do the following?

* remove the odoo version from the commit
* split the commit by module (guidelines [here](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message))
* rebase

PR #127, opened 2026-03-30
[19.0][MIG]: endpoint

Alexgars73:
Standard migration

Depends on
- https://github.com/OCA/server-tools/pull/3539


endpoint/models/endpoint_mixin.py
```diff
@@ -0,0 +1,274 @@
...
+        "new",
+    ],
+)
+hmac = safe_eval.wrap_module(
+    __import__("hmac"),
+    ["new", "compare_digest"],
+)
+
+
+@disable_rpc()  # Block ALL RPC calls
```
ivantodorovich, line 41: I would look into the new `@api.private` decorator in core: https://github.com/odoo/odoo/pull/195402

We may not need this `rpc_helper` dependency

⚠️ the `@api.private` must decorate methods, not classes. I'm not sure here what RPC methods we were trying to protect

ivantodorovich (approved):

yankinmax (approved):

endpoint/models/endpoint_mixin.py
Alexgars73: @ivantodorovich Thanks for the suggestion! I looked into @api.private but it doesn't behave the same way as @disable_rpc() for our use case.

@disable_rpc() patches execute_cr in odoo.service.model, which is the entry point only for XML-RPC (/xmlrpc/2/object). The Odoo web client takes a different path — it goes through /web/dataset/call_kw → call_kw() directly, bypassing execute_cr entirely. So the backend UI keeps working fine with @disable_rpc().

@api.private, on the other hand, is checked inside get_public_method() which is called by call_kw() — meaning it blocks both XML-RPC and the web client. If we decorated create/write/read etc. with @api.private, the backend views for managing endpoints would break.

ivantodorovich: To me it sounds like both attempt to do the same thing, but `@api.private` is complete and `@disable_rpc` is missing a case, but probably not intentionally.. cause I do not see the point in disabling a method in XML-RPC but not in the REST API.

Do you know which are the methods we want to disable RPC for in this module? If we understand the use case, we can discuss whether one solution fits or not.

cc @simahawk

grindtildeath: > If we decorated create/write/read etc. with @api.private, the backend views for managing endpoints would break.

IMO that's a legit reason to keep it as it is for now :+1:

grindtildeath (approved):

Reyes4711-S73 (approved):

endpoint/models/endpoint_mixin.py
simahawk: The main reason to block RPC calls was to avoid the creation or modification of records.
Hence, in this context, it could be enough to decorate all sensible methods w/ `api.private`:

- create
- write
- copy
- copy_data
- $what_else?

endpoint/models/endpoint_mixin.py
```diff
@@ -0,0 +1,274 @@
...
+        for rec in self:
+            rec.code_snippet_docs = textwrap.dedent(rec._default_code_snippet_docs())
+
+    @api.constrains("exec_mode")
+    def _check_exec_mode(self):
+        for rec in self:
+            rec._validate_exec_mode()
+
+    def _validate_exec_mode(self):
+        validator = getattr(self, "_validate_exec__" + self.exec_mode, lambda x: True)
```
ivantodorovich:
```suggestion
        validator = getattr(self, "_validate_exec__" + self.exec_mode, lambda: True)
```

the validator is called without arguments, so the default lambda value shuoldn't take any

ivantodorovich: this is a pre-existing bug. can you solve in a separate commit so we can easily backport?

yankinmax:
Hello @Alexgars73 , will you be able to update and fix this PR soon or I can replace it with another one and speed up the migration process?

Alexgars73:
@yankinmax i added the change that @ivantodorovich request, the dependency still not merged so we cannot do nothing at moment.

yankinmax:
> @yankinmax i added the change that @ivantodorovich request, the dependency still not merged so we cannot do nothing at moment.

Thanks a lot. I'll try to speed up `rpc_helper` migration

yankinmax:
Hey @Alexgars73 the `rpc_helper` is merged, let's update this one and get it merged :smile:

Alexgars73:
@yankinmax done!

yankinmax:
Hello @simahawk can you pls trigger a merge here?

simahawk (approved):

PR #126, opened 2026-03-24
[19.0][MIG] webservice and webservice_server_env: Migration to v19

AaronHForgeFlow:
Based on https://github.com/OCA/web-api/pull/101

PR #128, opened 2026-04-01
[19.0][MIG] webservice and webservice_server_env: Migration to v19

yankinmax:
Standard migration.

Depends on:
- [ ] https://github.com/OCA/web-api/pull/126

1. This PR is opened to speed up migration to 19.0.
2. Fixed tests and squashed administrative commits.

cc @AaronHForgeFlow


webservice/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "WebService",
+    "summary": """Defines webservice abstract definition to be used generally""",
+    "version": "19.0.1.1.0",
+    "license": "AGPL-3",
+    "development_status": "Production/Stable",
+    "maintainers": ["etobella"],
+    "author": "Creu Blanca, Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web-api",
+    "depends": ["component", "web"],
```
yankinmax: `web` is added to dependencies, because we want to test some UI flow and use `Form` in the tests.
To do it we need `web` dependency:
```
ERROR odoo odoo.addons.webservice.tests.test_oauth2: ERROR: TestWebServiceOauth2WebApplication.test_oauth2_flow_compute_with_ui
Traceback (most recent call last):
  File "/__w/web-api/web-api/webservice/tests/test_oauth2.py", line 233, in test_oauth2_flow_compute_with_ui
    with Form(ws.browse(), form_xmlid) as ws_form:
  File "/opt/odoo/odoo/tests/form.py", line 143, in __init__
    self._init_from_defaults()
  File "/opt/odoo/odoo/tests/form.py", line 339, in _init_from_defaults
    self._perform_onchange()
  File "/opt/odoo/odoo/tests/form.py", line 579, in _perform_onchange
    result = record.onchange(values, field_names, self._view['fields_spec'])
  File "/opt/odoo/odoo/orm/models.py", line 6994, in onchange
    raise NotImplementedError("onchange() is implemented in module 'web'")
NotImplementedError: onchange() is implemented in module 'web'
```

ivantodorovich (approved):

AaronHForgeFlow (approved):
Thanks

webservice/__manifest__.py
grindtildeath: Looking at the [original commit](https://github.com/camptocamp/web-api/commit/44993363ce481c4a141d7b8e9bccfa3c827db767) where the use of `Form` was added in the tests, my feeling is it was done to test how the UI behaves with server-env installed but values for that record are not defined in server-env (since it's actually testing something that what was set in the test and not any compute or onchange function). 

I would avoid adding `web` as a dependency to such a technical module only to run tests, so I'd suggest to move said test to the `webservice_server_env` module.

grindtildeath (changes requested):
IMO we shouldn't depend on web for tests, cf my comment

setup/webservice_server_env/odoo/addons/webservice_server_env
```diff
@@ -0,0 +1 @@
+../../../../webservice_server_env
```
vvrossem, line 1: missing EOL

yankinmax: This file shouldn't be here in v19. Thanks for a good catch

yankinmax:
Hello @OCA/board can you pls take a look and possibly merge?

PR #131, opened 2026-04-30
[19.0][IMP] endpoint: request content schema

ivantodorovich:
Implement request content validation through an optional schema.

For `application/json`, we use the JSON Schema (same than openapi)
For `application/xml`, we use XSD Schema

### DIFF
--- a/endpoint/models/endpoint_mixin.py
+++ b/endpoint/models/endpoint_mixin.py
@@ -2,15 +2,21 @@
 # @author: Simone Orsi <simone.orsi@camptocamp.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+import json
 import textwrap
 
+import jsonschema
 import werkzeug
+import yaml
+from lxml import etree
 
 from odoo import api, exceptions, fields, http, models
 from odoo.exceptions import UserError
 from odoo.tools import safe_eval
 
 from odoo.addons.rpc_helper.decorator import disable_rpc
+
+from ..exceptions import RequestValidationError
 
 hashlib = safe_eval.wrap_module(
     __import__("hashlib"),
@@ -48,6 +54,18 @@ class EndpointMixin(models.AbstractModel):
         selection="_selection_exec_mode",
         required=True,
     )
+    request_content_schema = fields.Text(
+        help=(
+            "Optional schema validated against the parsed request body. "
+            "The accepted format depends on 'Request content type':\n"
+            " - application/json: JSON Schema (Draft 2020-12), as YAML or JSON\n"
+            " - application/xml, text/xml: XML Schema (XSD)\n"
+            "If empty, no validation runs."
+        ),
+    )
+    request_content_schema_applicable = fields.Boolean(
+        compute="_compute_request_content_schema_applicable",
+    )
     code_snippet = fields.Text()
     code_snippet_docs = fields.Text(
         compute="_compute_code_snippet_docs",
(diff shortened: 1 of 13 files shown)


endpoint/models/endpoint_mixin.py
```diff
@@ -80,6 +98,69 @@ def _validate_exec__code(self):
...
+            jsonschema.Draft202012Validator.check_schema(schema)
+        except jsonschema.SchemaError as exception:
+            raise UserError(
+                self.env._(
+                    "Invalid JSON Schema in request content schema: %s",
+                    exception.message,
+                )
+            ) from exception
+
+    def _check_request_content_schema_xml(self):
```
simahawk, line 147: We have plenty of XSD files in modules: we should be able to provide a path too.
If you don't want to add a specific field for this, we could use a prefix `path:`.

simahawk (changes requested):
README to be updated

endpoint/models/endpoint_mixin.py
ivantodorovich: What use case do you have in mind?

IMO if you want to specify the schema in a different file, it can be covered with standard Odoo xml import, like this. 

```xml
    <record id="endpoint_demo" model="endpoint.endpoint">
        <field name="name">Demo Endpoint</field>
        <field name="route">/demo</field>
        <field name="request_method">POST</field>
        <field name="exec_mode">code</field>
        <field name="code_snippet">result = {"response": Response("Hello, World!")}</field>
        <field name="request_content_type">application/xml</field>
        <field name="request_content_schema" type="char" file="endpoint/demo/content_schema.xsd" />
    </record>
```

ivantodorovich:
@simahawk I added the configure section in the readme

endpoint/models/endpoint_mixin.py
simahawk: that won't work from the UI.
The use case is: I have a URL to XSD or an attachment: how would I use it here w/o copy/pasting its content?
I would at list allow to select an existing attachment instead w/ domain filter on `*.xsd` files.

ivantodorovich: For a pure UI created record, I think it's not that problematic if the user copy&paste the content directly in the field. There's no benefit in having a separate attachment for this, in fact in practice it will make it slower due to binary reads than just having its content in the DB.

I rather not add a magic syntax like "path:" to the field content. IMO, the schema field content should be a pure schema definition, either XSD or JSON schema (or something else in the future for other content types), not a magic thing built on top.

So, if we were to do this cleanly, I'd go for a `request_content_schema_source` and separate fields for each source type. But, honestly, it's seems overly complicated to avoid a copy&paste...

endpoint/models/endpoint_mixin.py
```diff
@@ -80,6 +98,69 @@ def _validate_exec__code(self):
...
+            rec.request_content_schema_applicable = (
+                rec.request_method in ("POST", "PUT")
+                and rec.request_content_type in applicable_types
+            )
+
+    @api.onchange("request_content_type")
+    def _onchange_request_content_type_clear_schema(self):
+        # The schema format depends on the content type (e.g. JSON Schema vs
+        # XSD), so it cannot survive a content type change.
+        self.request_content_schema = False
```
Ricardoalso, line 118:
```suggestion
        self.request_content_schema = ""
```

Nit: I personally find it more self-explanatory

Ricardoalso (approved):
Technically it LGTM

Thanks 🙏

vvrossem (approved):
🐐

endpoint/models/endpoint_mixin.py
simahawk: I disagree :) 

With this approach when you have N endpoints requiring the same schema (I have cases like that) you have to copy/paste all the times you have a little change. Not to mention the fact that you might bloat the DB in that case.

I'm not blocking this early version but could you please add a TODO to follow up w/ a better solution?

ivantodorovich: Ok for the TODO :) 

But seriously, would you maintain these endpoints through the UI and not by code? because if you do by code you can use the `type="char" file=".."` attributes I mentioned before, and point to the shared schema file

endpoint/models/endpoint_mixin.py
ivantodorovich: It's not the same thing, though.

`False` results in `NULL` in the db, whilst `""` is an empty string.

endpoint/models/endpoint_mixin.py
ivantodorovich: Roadmap added

endpoint/static/description/index.html
```diff
@@ -385,54 +385,195 @@ <h1>Endpoint</h1>
...
 <div class="contents local topic" id="contents">
 <ul class="simple">
-<li><a class="reference internal" href="#configuration" id="toc-entry-1">Configuration</a></li>
-<li><a class="reference internal" href="#known-issues-roadmap" id="toc-entry-2">Known issues / Roadmap</a></li>
-<li><a class="reference internal" href="#bug-tracker" id="toc-entry-3">Bug Tracker</a></li>
-<li><a class="reference internal" href="#credits" id="toc-entry-4">Credits</a><ul>
-<li><a class="reference internal" href="#authors" id="toc-entry-5">Authors</a></li>
-<li><a class="reference internal" href="#contributors" id="toc-entry-6">Contributors</a></li>
-<li><a class="reference internal" href="#maintainers" id="toc-entry-7">Maintainers</a></li>
+<li><a class="reference internal" href="#configuration" id="toc-entry-1">Configuration</a><ul>
```
simahawk, line 388: `export SKIP=oca-gen-addon-readme`

Next time 😉

simahawk (approved):

PR #133, opened 2026-05-21
[19.0] [IMP] endpoint: support cors setting

ivantodorovich:
This commits adds support for CORS setting on endpoints.

Without this, the default behavior is to disallow all origins. That's usually the opposite of what you want for an API endpoint, as you want consumers to be able to access the endpoint from their own origin.

Now, by default, the endpoint will allow all origins (default="*"), and it can be further configured for more specific restrictions.


simahawk (approved):

yankinmax (approved):

simahawk:
/ocabot merge minor
@ivantodorovich can you backport to 18 pls?

PR #135, opened 2026-05-26
[19.0][ADD] endpoint_json2

AungKoKoLin1997:
This module adds `exec_mode="json2 `to the endpoint framework, enabling declarative JSON-2 API endpoint configuration. Select a model, method, and parameters — the module handles dispatch, parameter validation, access control, and result filtering. A code snippet can be used as an alternative to a model method for quick, ad-hoc logic.

@qrtl QT6769


simahawk:
@AungKoKoLin1997 interesting approach. Early feedback: I don't think the mixin is the right place as it can be inherited by all extending models (eg: edi_endpoint). I tend to say `endpoint.endpoint` is better.

AungKoKoLin1997:
@simahawk Thanks for your feedback. I updated to use `endpoint.endpoint` instead of the mixin model.

endpoint_json2/readme/CONFIGURE.md
```diff
@@ -0,0 +1,54 @@
...
+
+All endpoint execution is wrapped in `sudo()`, allowing API users to operate with
+minimal Odoo privileges. Access is controlled at two levels:
+
+- **Auth Type**: Select the authentication method for the endpoint (e.g. **Bearer**
+  for API key authentication).
+- **Allowed Groups**: Restrict endpoint access to specific user groups. Create
+  integration-specific groups (e.g. "Hospital System", "WMS") and assign them to
+  the corresponding API users. Each endpoint declares which groups may call it.
+  Leave empty to allow any authenticated user.
```
smorita7749: We cannot leave Allowed Groups empty due to a UI restriction. 
Is it necessary to mention this?

smorita7749: This part:  Leave empty to allow any authenticated user.

smorita7749 (approved):
Functional review: LGTM
The flow creating endpoint and getting response is properly working.

simahawk:
@ivantodorovich can you have a look at this?

endpoint_json2/tests/common.py
```diff
@@ -0,0 +1,44 @@
...
+class CommonEndpointJson2(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.model_partner = cls.env["ir.model"]._get("res.partner")
+        cls.group_user = cls.env.ref("base.group_user")
+        cls.endpoint = cls._create_endpoint(
+            {
+                "name": "get_partners",
+                "route_group": "test_contacts",
```
yostashiro, line 18: I just updated this from `contacts` to avoid conflicts with the demo data.

endpoint_json2/models/endpoint_endpoint.py
```diff
@@ -0,0 +1,418 @@
...
+        default="[]",
+        help="Default domain filter applied before calling the method (JSON format).",
+    )
+    json2_lang_id = fields.Many2one(
+        "res.lang",
+        string="Response Language",
+        help="Force this language on the execution context so that translatable "
+        "field values (including dotted relational fields) are returned in it, "
+        "regardless of the API user's language.",
+    )
```
yostashiro, line 65: Added this field for convenience.

yostashiro (approved):
Code review and functional test.

AungKoKoLin1997:
I added a few changes.

- **Response Language** (`json2_lang_id`): optionally forces the execution context language, so translated values (including dotted relational fields like `uom_id.name`) no longer depend on the API user's language setting.
- **Response Timezone** (`json2_tz`): optionally converts datetimes in the response to the given timezone, rendered as local wall time without an offset (incoming datetime parameters are left as-is).
- **Align access / exec-target rules with the documentation**: an empty Allowed Groups list rejects every caller, and Method / Code Snippet are truly mutually exclusive — both now backed by constraints so a data file or an import cannot create an endpoint that nobody can call, or one with both exec targets set.
- **Response fields for non-read methods**: plain field names are validated only for `read` / `search_read`, whose rows are made of model fields. Other methods can now declare their own keys instead of leaving `json2_response_fields` empty (which disables filtering and aliasing entirely). Dotted specs stay validated for every method.
- **Tests**: moved the test endpoints to their own route group so they no longer collide with the demo data route.

yostashiro:
This PR now depends on https://github.com/OCA/web-api/pull/160.

endpoint_json2/controllers/main.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class EndpointJson2DocController(http.Controller):
+    def _get_accessible_endpoints(self, extra_domain=None):
+        domain = [("exec_mode", "=", "json2")] + (extra_domain or [])
+        all_endpoints = request.env["endpoint.endpoint"].sudo().search(domain)
+        user = request.env.user
+        return all_endpoints.filtered(
+            lambda ep: ep.json2_group_ids & user.all_group_ids
+        )
```
SilvioC2C, line 17: Nitpicking: use Odoo 19.0 syntax for domains
```suggestion
from odoo import http
from odoo.http import request
from odoo.orm.domains import Domain


class EndpointJson2DocController(http.Controller):
    def _get_accessible_endpoints(self, extra_domain: Domain | None = None):
        domain = Domain("exec_mode", "=", "json2")
        if extra_domain:
            domain &= extra_domain
        all_endpoints = request.env["endpoint.endpoint"].sudo().search(domain)
        user = request.env.user
        return all_endpoints.filtered(
            lambda ep: ep.json2_group_ids & user.all_group_ids
        )
...

File: endpoint_json2/controllers/main.py [hunk 5]
```diff
@@ -0,0 +1,67 @@
...
+    @http.route(
+        "/json2/doc/<string:route_group>",
+        methods=["GET"],
+        auth="user",
+        type="http",
+        readonly=True,
+        save_session=False,
+    )
+    def doc_domain(self, route_group):
+        endpoints = self._get_accessible_endpoints([("route_group", "=", route_group)])
```
SilvioC2C, line 62: Same comment as above about Odoo 19.0 domains' syntax

endpoint_json2/models/endpoint_endpoint.py
```diff
@@ -0,0 +1,464 @@
...
+    def _json2_is_valid_response_field(self, Model, field_spec):
+        if "." not in field_spec:
+            return field_spec in Model._fields
+        base, sub = field_spec.split(".", 1)
+        fd = Model._fields.get(base)
+        return (
+            fd
+            and fd.type in ("many2one", "many2many", "one2many")
+            and sub in self.env[fd.comodel_name]._fields
+        )
```
SilvioC2C, line 202: I think we can improve this method to:
1- not need the `Model` param, or make it non-mandatory
2- allow deeper dotted fields (e.g.: `sale_id.commercial_partner_id.name`)

```suggestion
    def _json2_is_valid_response_field(self, field_spec, model=None):
        if model is None:
            model = self.env[self.json2_model_name]
        model = model.sudo()  # Prevent ``AccessError``
        for field_name in field_spec.split("."):
            if field_name not in model._fields:
                return False
            model = model[field_name]
        return True
```

endpoint_json2/models/endpoint_endpoint.py
```diff
@@ -0,0 +1,464 @@
...
+            params[param_def.name] = param_def._extract_value(
+                kwargs.pop(param_def.name, None)
+            )
+        return params
+
+    def _json2_get_code_snippet_eval_context(self, Model, params):
+        return {
+            "Model": Model,
+            "params": params,
+            "env": Model.env,
```
SilvioC2C, line 257:
```suggestion
    def _json2_get_code_snippet_eval_context(self, params):
        return {
            "Model": self.env[self.json2_model_name],
            "params": params,
            "env": self.env,
```

SilvioC2C (approved):
Very nice addition, thanks!

A few possible improvements:
- allow dotted fields to go deeper than 1 model (e.g.: `sale_id.commercial_partner_id.name`)
- the `Model` param in methods' signatures could be dropped, or passed as an optional param (i.e.: `Model=None`) to force the model instead of fetching it from `self`
- use Odoo 19.0 `Domain` syntax

endpoint_json2/controllers/main.py
AungKoKoLin1997: Done!

endpoint_json2/controllers/main.py
AungKoKoLin1997: Done!

endpoint_json2/models/endpoint_endpoint.py
AungKoKoLin1997: Thanks, applied the change.

I kept _fields instead of model[field_name]. On an empty recordset a scalar field returns False, so email.something would raise AttributeError instead of a ValidationError (there is a test for that input).

endpoint_json2/models/endpoint_endpoint.py
AungKoKoLin1997: I would prefer to keep Model here.

It is not the same as self.env[self.json2_model_name]. In _handle_exec__json2 it is request.env[...].sudo() with the lang and tz context applied.

Code snippets are meant to run with sudo, because access is controlled by json2_group_ids and not by the model's access rights. And json2_lang_id / json2_tz need to reach them. Taking the model from self would lose all three.

PR #136, opened 2026-05-26
[19.0][FIX] endpoint_route_handler: precompute=True on route field

AungKoKoLin1997:
Related: https://github.com/OCA/web-api/pull/135

Without precompute, the required computed field is evaluated after INSERT (during flush), so creating an endpoint whose route is derived from other fields by a downstream module fails with "Missing required value for the field 'Route'" — the INSERT runs before the compute can populate the column.

@qrtl QT6769

### DIFF
--- a/endpoint_route_handler/tests/test_endpoint.py
+++ b/endpoint_route_handler/tests/test_endpoint.py
@@ -2,6 +2,7 @@
 # @author: Simone Orsi <simone.orsi@camptocamp.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 from contextlib import contextmanager
+from unittest.mock import patch
 
 from odoo import api, modules
 from odoo.tests import common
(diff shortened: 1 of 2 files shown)


yostashiro (approved):
Code review. LGTM.

smorita7749 (approved):
Functional review: LGTM
The error is fixed.

endpoint_route_handler/tests/test_endpoint.py
```diff
@@ -50,6 +51,33 @@ def test_as_tool_base_data(self):
         new_route.route += "/new"
         self.assertNotEqual(new_route.endpoint_hash, first_hash)
 
+    def test_route_field_precomputed(self):
```
simahawk, line 54: do we really need this test? It's kind of testing an odoo core feature 🤔

AungKoKoLin1997: My purpose is to be sure the fix is working well.

AungKoKoLin1997: @simahawk 
To clarify further — the test is not meant to verify Odoo's precompute mechanism itself. It's a regression guard to ensure `precompute=Tru`e stays on the route field. If someone removes it during a future migration or refactor, this test will catch the breakage. Without it, the only way to notice would be a failure in a downstream module at runtime.

simahawk: ok, you could rewrite the docstring of the test by removing repeated information (sounds quite verbose ATM for little value) and add a synthetic version of the purpose that you've just explained.

AungKoKoLin1997: @simahawk I updated the docstring.

simahawk (approved):
@AungKoKoLin1997 can you backport to v18 pls?

AungKoKoLin1997:
> can you backport to v18 pls?

@simahawk Done!
https://github.com/OCA/web-api/pull/165

PR #138, opened 2026-05-29
[19.0][FIX] webservice: allow to return response object on OAuth2-configure…

Ricardoalso:
…d calls

When sending exchanges through an OAuth2-configured webservice.backend, the request crashed with:

`TypeError: Session.request() got an unexpected keyword argument 'content_only'
`

Stack trace pointed to webservice/components/request_adapter.py in BackendApplicationOAuth2RestRequestsAdapter._request, where content_only was forwarded to OAuth2Session.request(...)

Also check https://github.com/OCA/web-api/pull/48

-----

`

**Open question** :warning: 


ivantodorovich (approved):

simahawk (approved):

simahawk:
/ocabot merge patch

@Ricardoalso can you pls backport it to 18?

PR #140, opened 2026-06-04
[19.0][IMP][WIP] endpoint_route_handler: memoize route version per request

JordiBForgeFlow:
_endpoint_route_last_version() is part of the routing_map ormcache key, so it runs on every routing map access -- e.g. once per url_for while rendering a website page, which adds up to dozens of identical 'SELECT last_value FROM endpoint_route_version' round-trips per page on content-heavy sites. The version cannot change within a request: memoize it on the request object.

### DIFF
--- a/endpoint_route_handler/models/ir_http.py
+++ b/endpoint_route_handler/models/ir_http.py
@@ -43,10 +43,44 @@ class IrHttp(models.AbstractModel):
         res = super().routing_map(key=key)
         return res
 
+    # Sentinel stored in the request `__dict__` to memoize the version.
+    _endpoint_route_version_attr = "_endpoint_route_last_version"
+
     @classmethod
     def _endpoint_route_last_version(cls):
-        res = cls._get_routing_map_last_version(http.request.env)
-        return res
+        # This is part of the `routing_map` ormcache key, so it runs on
+        # every routing map access (e.g. once per `url_for` while rendering
+        # a website page). The version only changes when routes are
+        # (un)registered, which resets the memo via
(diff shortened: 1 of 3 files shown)


endpoint_route_handler/models/ir_http.py
```diff
@@ -43,10 +43,44 @@ def routing_map(self, key=None):
...
+        # object to avoid one SQL round-trip per call.
+        # NB: use `__dict__` rather than `getattr`: on a mocked request
+        # `getattr` would auto-create a child mock instead of falling back
+        # to the default, defeating the memoization.
+        # NB: `http.request` is a werkzeug `LocalProxy`; when no request is
+        # bound it is falsy (but not `None`), hence the truthiness check.
+        request = http.request
+        if not request:
+            return cls._get_routing_map_last_version(http.request.env)
+        version = request.__dict__.get(cls._endpoint_route_version_attr)
```
simahawk, line 65: are we sure there's no better way than accessing `__dict__`? I tend to think the mocking is wrong somehow.

You could deleted it beforehand:
```
>>> from unittest.mock import MagicMock
>>> my_mock = MagicMock()
>>> del my_mock.missing_attr
>>> getattr(my_mock, "missing_attr", "default")
'default'
>>> getattr(my_mock, "missing_attr", "default")
'default'
>>> my_mock.missing_attr = 1
>>> getattr(my_mock, "missing_attr", "default")
1
```
Is that what you are looking for?

simahawk:
Note: this will clash with https://github.com/OCA/web-api/pull/159

PR #143, opened 2026-07-02
[19.0] [IMP] webservice: configurable OAuth2 token request

ivantodorovich:
Some providers require the credentials in a non-standard Authorization header (for instance Okta uses `Authorization: SSWS <token>`).

This PR makes it possible. Such an endpoint can be configured as:

    Auth Type             = OAuth2
    OAuth2 Flow           = Backend Application (Client Credentials Grant)
    Token URL             = https://provider.example.com/oauth2/token
    Client Authentication = Custom Authorization header
    Client Auth Header    = Authorization
    Client Auth Value     = SSWS <token>

Additionally, the provider might only reply to `GET` instead of `POST`, so the token request method is configurable now too.

### DIFF
--- a/webservice/views/webservice_backend.xml
+++ b/webservice/views/webservice_backend.xml
@@ -61,7 +61,7 @@
                         />
                         <field
                             name="redirect_url"
-                            invisible="auth_type != 'oauth2' and oauth2_flow != 'web_application'"
+                            invisible="auth_type != 'oauth2' or oauth2_flow != 'web_application'"
                         />
                         <field
                             name="oauth2_client_auth_method"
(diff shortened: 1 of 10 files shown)


Ricardoalso (approved):

SilvioC2C (approved):

ivantodorovich:
Hello @etobella 👋🏻

What do you think about this?

webservice/views/webservice_backend.xml
```diff
@@ -61,33 +61,56 @@
...
+                        <field
+                            name="oauth2_client_auth_method"
+                            invisible="auth_type != 'oauth2' or oauth2_flow != 'backend_application'"
+                            widget="radio"
                         />
                         <field
                             name="oauth2_clientid"
-                            invisible="auth_type != 'oauth2'"
-                            required="auth_type == 'oauth2'"
+                            invisible="auth_type != 'oauth2' or (oauth2_flow == 'backend_application' and oauth2_client_auth_method != 'client_secret_basic')"
```
simahawk, line 73: wouldn't be better to have a flag to compute this condition? Eg: `oauth2_clientid_is_visible`

simahawk (commented):
LG overall. Minor remark on form modifiers

webservice/views/webservice_backend.xml
ivantodorovich: Not sure.

We'd need at least two fields:
`oauth2_client_secret_fields_visible` (for `oauth2_clientid` + `oauth2_client_secret`)
`oauth2_client_custom_header_fields_visible` (for `oauth2_client_auth_header` + `oauth2_client_auth_value`)

Personally, I'm more inclined to keep them inline in the view, and avoid paying the cost of the dedicated computed fields (a backend request roundtrip) -- IMHO it's not complex enough.. but I'm willing to change if you disagree

simahawk: not a blocker. go ahead ;)

PR #144, opened 2026-07-04
[FIX] webservice: deprecate implicit content_only=True default

### LINKED ISSUE #142: Open Concern about webservice module defaulting content_only to True in "base.requests"
## Module
webservice

## Describe the bug
[Defaulting content_only to True in "base.requests"](https://github.com/OCA/web-api/blob/18.0/webservice/components/request_adapter.py#L29) seems wrong IMHO.
A standard call to an endpoint self._request("GET", "/test") returns raw bytes instead of a typical HTTP response object

## To Reproduce
Call _request(self, method, url=None, url_params=None, **kwargs)

**Expected behavior**
When calling _request we should get an HTTP response object

**Additional context**
Check originating PR https://github.com/OCA/web-api/pull/138 with the "Open Concern" section 

<details>
<summary>Click to expand original PR content</summary>

`

**Open question** :warning: 

</details>

@gurneyalex @simahawk @etobella

HasanAlsaafen:
Fixes #142

## Problem
`BaseRestRequestsAdapter._request()` defaults `content_only` to `True`,
returning raw `bytes` instead of the expected `requests.Response` object.
This violates the Principle of Least Astonishment, as most HTTP client
libraries return a Response object by default.

Additionally, `BackendApplicationOAuth2RestRequestsAdapter._request()`
ignored `content_only` entirely, always returning `.content`, which was
inconsistent with the base adapter.

## Solution
- `content_only` now defaults to `None`. When not explicitly set, a
  deprecation warning is logged and the current behavior (`True`) is
  preserved for backward compatibility.
- Fixed the OAuth2 backend application adapter to respect `content_only`
  consistently with the base adapter.
- Added tests covering: default behavior with warning, explicit
  `content_only=True`, and explicit `content_only=False`.

## Testing
All 25 existing tests pass alongside the 3 new tests added.

### DIFF
--- a/webservice/components/request_adapter.py
+++ b/webservice/components/request_adapter.py
@@ -26,7 +26,14 @@ class BaseRestRequestsAdapter(Component):
     # TODO: url and url_params could come from work_ctx
     def _request(self, method, url=None, url_params=None, **kwargs):
         url = self._get_url(url=url, url_params=url_params)
-        content_only = kwargs.pop("content_only", True)
+        content_only = kwargs.pop("content_only", None)
+        if content_only is None:
+            _logger.warning(
+                "The default value of 'content_only' will change from "
+                "True to False in a future version. "
+                "Please set it explicitly to avoid unexpected behavior."
+            )
+            content_only = True
         # TODO: turn on/off debug from webservice setting?
         url_to_log = self._sanitize_url_for_log(url)
         _logger.info("%s call to %s", method, url_to_log)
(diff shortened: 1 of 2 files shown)


Ricardoalso (commented):
Hey @HasanAlsaafen, thanks for this PR related to https://github.com/OCA/web-api/issues/142.

I originally opened the issue to gather external input on how to handle this situation, if we decide to address it. The challenge with making changes like the ones introduced in this commit is that they may disrupt the expected behavior of existing modules or code that depend on the current default, potentially breaking running instances.

HasanAlsaafen:
Hey @Ricardoalso 
Thanks for raising this, it's a valid concern.

Before deciding how to proceed, it might help to first get a clearer picture of the actual impact: are there known modules in the OCA ecosystem that rely on the current default behavior, or would this mainly affect external/custom code outside OCA? If we can narrow down who is likely affected, it'll be easier to choose between an opt-in flag, a deprecation cycle, or a straight breaking change with a migration note.

Let me know if there's a preferred way to check this (e.g. searching OCA repos for usages), and I can help gather that data before we settle on an approach.

simahawk:
@HasanAlsaafen sounds like a good approach. Can you pls rewrite the commit msg as `[FIX] webservice: deprecate implicit content_only=True default`.
Regarding other modules: I don't know any other module in the OCA that directly relies on `webservice.call` but you could search for modules depending on it. 
Last but not least. pls check the CI :pray:

simahawk:
@Hasan2005-CS thanks for your contrib. In the end I had to choose a slightly different approach because of the core split.
It's here https://github.com/OCA/web-api/pull/167 if you want to have a look.

PR #145, opened 2026-07-09
[19.0][IMP] webservice, webservice_server_env: neutralize webservice backend credentials

yankinmax:


webservice/data/neutralize.sql
```diff
@@ -0,0 +1,8 @@
+-- remove webservice backend credentials
+UPDATE webservice_backend
+   SET username = NULL,
+       password = NULL,
+       api_key = NULL,
+       oauth2_clientid = NULL,
+       oauth2_client_secret = NULL,
```
Ricardoalso, line 7: After this change (https://github.com/OCA/web-api/pull/143/changes), we will also need to update `oauth2_client_auth_method = NULL`

webservice_server_env/data/neutralize.sql
```diff
@@ -0,0 +1,15 @@
+-- remove webservice backend credentials stored as server environment defaults
+-- (this module turns username/password/api_key/oauth2_clientid/oauth2_client_secret
+-- into non-stored fields backed by the server_env_defaults JSON column
+-- this neutralization is needed if both webservice and server_environment are installed,
+-- so the plain columns nulled by webservice/data/neutralize.sql can be properly cleared)
+UPDATE webservice_backend
```
Ricardoalso, line 6: Same comment as above

Ricardoalso (commented):
Thanks @yankinmax 🙏 

Comments related to an open PR that will conflict

webservice/data/neutralize.sql
yankinmax: I've added this dependency

SilvioC2C: @yankinmax since this PR now depends on #143, you could also update the code to add the necessary changes to handle [these fields](https://github.com/OCA/web-api/pull/143/changes#diff-bb9b1cc04ef29b27e8d7de3e85370a608b6c8cac89e3f9376e8e80b83327a800R32-R35)?

SilvioC2C (commented):
Needs updating on top of #143

ivantodorovich (approved):
Thanks!

ivantodorovich:
@SilvioC2C @Ricardoalso this should be ok now, can you update your reviews? 🙏🏻

Ricardoalso (approved):

ivantodorovich:
@etobella / @simahawk can we merge this one? 🙏🏻

SilvioC2C:
@ivantodorovich hello, sorry but I was on vacation.
From what I can see, some of the new fields introduced by #143 are not neutralized by the `.sql` files' scripts. Is this wanted?

ivantodorovich:
> @ivantodorovich hello, sorry but I was on vacation. From what I can see, some of the new fields introduced by #143 are not neutralized by the `.sql` files' scripts. Is this wanted?

The missing fields are:
- oauth2_client_auth_method
- oauth2_client_auth_header
- oauth2_token_method

These don't hold any secrets, so IMO it's ok.
The only added field that would store a secret is: oauth2_client_auth_value, and that's on the list

SilvioC2C (approved):
@ivantodorovich thanks for the explanation, full green light from me then :green_circle:

simahawk (approved):

PR #141, opened 2026-06-30
[19.0] [IMP] `webservice_server_env`: uninstall hook

ivantodorovich:
**Depends on:**
- https://github.com/OCA/server-env/pull/274

PR #146, opened 2026-07-09
[18.0][ADD] webservice_server_env: make server_environment an optional dependency

yankinmax:
This PR is opened to speed up and replace work started on this PR:
 - [ ] https://github.com/OCA/web-api/pull/101

Applied @ivantodorovich request:
- [ ] https://github.com/OCA/web-api/pull/101#discussion_r3111416185

PR is aimed to remove `server_environment` dependency from `webservice` in the scope of this issue:
- [ ] https://github.com/OCA/server-env/issues/258

Backport `server_environment` uninstall hook:
- [x] https://github.com/OCA/web-api/pull/141

Depends on:
- [x] https://github.com/OCA/web-api/pull/148
- [x] https://github.com/OCA/server-env/pull/274


yankinmax:
Hello @dreispt @ivantodorovich @simahawk  can you pls take a look?

ivantodorovich (approved):

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-endpoint @ git+https://github.com/OCA/web-api.git@refs/pull/148/head#subdirectory=endpoint
```
ivantodorovich: This is merged 🚀 
you can drop the test requirement 

- https://github.com/OCA/web-api/pull/148

ivantodorovich (approved):

ivantodorovich:
@simahawk @dreispt

yankinmax:
@ivantodorovich I've updated this PR with backport commit of uninstall hook to be consistent with v19

yankinmax:
Hello @simahawk @etobella @dreispt , can you pls take a look and possibly trigger merge?

dreispt (approved):

webservice/upgrades/18.0.1.1.1/post-update.py
```diff
@@ -0,0 +1,13 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    module = env["ir.module.module"].search(
+        [("name", "=", "webservice_server_env"), ("state", "=", "uninstalled")]
+    )
+    if module:
+        module.button_install()
```
etobella, line 22: in the case that the module is not added, can you provide the uninstall hook to pass all data?

yankinmax: The uninstall hook is here: b175404472073c5b298851b92d9bdb7181a95dd3

etobella: in the migration should be there if the other module is not installed (it does not exist)

in some cases that could happen if you only add the modules

yankinmax: What do you mean?
In general we should have two cases:
1. `webservice` is already installed in the database (meaning current database version of the module depends on `server_environment`): then `webservice` upgrade will remove the dependency and upgrade hook will install `webservice_server_env`. 
2. `webservice` is not installed and this is a brand new installation: if `server_environment` is installed then `webservice_server_env` will be auto installed, because of `auto_install` key in manifest.
In both cases if you decide to uninstall `server_environment` and `webservice_server_env` the uninstall hook will make the fields regular for the `webservice`.

Can you pls describe the case I need to handle?

etobella: No, there is a 3rd case (and it happened to me with fs_storage).

You have server_environment and webservice installed. On the update, you don't know that you need the new module, and you don't add it (I use doodba, and I specify exactly the modules I need). On the update, the if is false, as the module is not found, then, you loose all data, leaving your instance without the credentials.

We have 2 options:

1- raise an exception if the module is not found
2- Make it in a way that the module restores data in case the module is not found.

yankinmax: > You have server_environment and webservice installed. On the update, you don't know that you need the new module, and you don't add it

I thought it's actually covered by Odoo with `auto_install`: if you potentially specify the list of modules to install, the `web-api` submodule is present on a branch revision (or even aggregated one is built on top of the branch revision), so you definitely have in the code source the `webservice_server_env`, then even if you don't force it in the list it should be auto installed.
Am I not right?

@ivantodorovich @simahawk @gurneyalex WDYT?

etobella: Yes, but that implies that the module is there. A lot of people is adding the whole branch of modules of the OCA repository, but that is not mandatory. That happened to me, with a system that uses git-aggregate and a process to get only the necessary modules. That could happen to if you use PyPi too. Doing this kind of activities (limiting the modules) allows us to be sure that the modules in the instance are compatible and allows us to make some tests on the instance before pushing it.

yankinmax: ok, I see....but I still can't get what kind of uninstall hook I need to add. Can you pls suggest the pseudo-code idea?

etobella: Something like this:

https://github.com/OCA/web-api/pull/146/changes#diff-477d83525842ed4ad9af6ca90b3a4d7c434ed43d33a48065d35662c006445572

Another option is to raise an Exception, this way this problem is avoided.

yankinmax: @etobella WDYT:
```
def migrate(env, version):
    module = env["ir.module.module"].search([("name", "=", "webservice_server_env")])
    if not module:
        raise exceptions.UserError(
            _(
                "The 'webservice_server_env' module is not available. "
                "It is required to preserve the server environment managed "
                "fields of 'webservice.backend'. Make it available on the "
                "addons path before upgrading 'webservice'."
            )
        )
    if module.state == "uninstalled":
        module.button_install()
```

simahawk: @etobella ping :)

etobella (approved):
Perfect! Thanks!

PR #132, opened 2026-05-19
[18.0] endpoint: fix lambda

ivantodorovich:
Backport from the 19.0 migration

PR #148, opened 2026-07-09
[FIX] endpoint: use self.patch for _endpoint_route_prefix in test_routing

yankinmax:
Manually reassigning type(endpoint)._endpoint_route_prefix leaves a residual class attribute directly on endpoint.endpoint (the field is only defined on the EndpointRouteHandler mixin), which Odoo's test harness now flags as an unexpected class mutation after the test. Use self.patch() so the attribute is properly restored (deleted) on cleanup.

### DIFF
--- a/endpoint/tests/test_endpoint.py
+++ b/endpoint/tests/test_endpoint.py
@@ -168,7 +168,7 @@ class TestEndpoint(CommonEndpoint):
             },
         )
         # check prefix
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):
@simahawk this is probably what fixes https://github.com/OCA/web-api/pull/132 too

simahawk (approved):

PR #150, opened 2026-08-17
[19.0][Fix]`endpoint_route_handler`: Restore registry._init_modules after mocked request

Camille0907:
`_get_mocked_request()` sets `registry._init_modules = set()` to simulate a minimal endpoint environment. Since registry is the real Odoo Registry singleton, this mutation persisted after the context exited.

Odoo core `ir_http.py` uses `registry._init_modules` to build the routing map, so leaving it as set() caused post_install HttpCase tests from other modules to receive a routing map containing only server-wide modules, resulting in 404 errors on routes registered by those modules.

Fix by saving and restoring _init_modules in a try/finally block.

### DIFF
--- a/endpoint_route_handler/tests/common.py
+++ b/endpoint_route_handler/tests/common.py
@@ -38,6 +38,8 @@ class CommonEndpoint(TransactionCase):
     def _get_mocked_request(
         self, env=None, httprequest=None, extra_headers=None, request_attrs=None
     ):
+        registry = (env or self.env).registry
+        original_init_modules = registry._init_modules
         with MockRequest(env or self.env) as mocked_request:
             mocked_request.httprequest = (
                 DotDict(httprequest) if httprequest else mocked_request.httprequest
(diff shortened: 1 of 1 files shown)


endpoint_route_handler/tests/common.py
```diff
@@ -38,16 +38,24 @@ def _setup_records(cls):
...
-            mocked_request.httprequest.headers = headers
-            request_attrs = request_attrs or {}
-            for k, v in request_attrs.items():
-                setattr(mocked_request, k, v)
-            mocked_request.make_response = lambda data, **kw: data
-            mocked_request.registry._init_modules = set()
-            yield mocked_request
+        registry = (env or self.env).registry
+        original_init_modules = registry._init_modules
+        try:
```
simahawk: IMO there's no need for a try/finally.
This is ctx manager, you can cleanup what you want after the yield statement.

simahawk (changes requested):
Thanks for your fix.
Please, amend the commit msg to remove the odoo version and use the tag `[FIX]`

simahawk (approved):

simahawk:
/ocabot merge patch
@Camille0907 can you pls backport to 18?

PR #152, opened 2026-08-18
[19.0][FIX] endpoint: validate registry sync

le-goff-benoit:
## What changed

- validate active endpoint records before scheduling a registry synchronization
- compile dynamic routes with the exact converters exposed by `ir.http`
- validate endpoint snippets against `safe_eval` syntax and detect unavailable global variables
- derive accepted variables from the actual evaluation context and Odoo built-ins
- keep invalid endpoints editable while preventing them from reaching `endpoint_route`
- keep invalid inactive endpoints synchronizable so stale registry rules can be removed
- repeat validation in controller registration methods as a defense for programmatic callers

## Why

Werkzeug compiles every registered URL rule while Odoo builds its global routing map. A dynamic endpoint using an unknown converter, such as `<date:value>`, raises a `LookupError` and prevents the entire routing map from being built. This makes every application route return an error, not only the invalid endpoint.

Endpoint code snippets can also reference names that are not part of the `safe_eval` context. For example, `date(...)` is not available, while `datetime` is.

The endpoint model already separates editing from publication through the `registry_sync` action. This change uses that boundary to report actionable errors without preventing users from saving an invalid draft.

## User impact

When an active invalid endpoint is synchronized, the user now receives a `UserError` listing the affected endpoint and its route or code errors. No invalid rule is written to the registry. After correcting the endpoint, the same synchronization action publishes it normally.

## Validation

- OCA pre-commit suite: passed
- Odoo 19.0 CI image (`py3.10-odoo19.0`): 102 tests, 0 failures, 0 errors
- regression tests cover unknown and malformed route converters, invalid `safe_eval` syntax, unavailable system variables, accepted built-ins, draft persistence, successful synchronization after correction, and removal of an invalid inactive route

### DIFF
--- a/endpoint/models/endpoint_mixin.py
+++ b/endpoint/models/endpoint_mixin.py
@@ -3,6 +3,7 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import json
+import symtable
 import textwrap
 
 import jsonschema
@@ -42,6 +43,10 @@ hmac = safe_eval.wrap_module(
     __import__("hmac"),
     ["new", "compare_digest"],
 )
+
+# ``safe_eval`` injects these built-ins at runtime. Keep validation aligned with
+# Odoo instead of maintaining a second, potentially outdated list here.
+_SAFE_EVAL_BUILTINS = frozenset(safe_eval._BUILTINS)
 
 
 @disable_rpc()  # Block ALL RPC calls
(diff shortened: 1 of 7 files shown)


endpoint/models/endpoint_mixin.py
```diff
@@ -98,6 +103,62 @@ def _validate_exec__code(self):
                 )
             )
 
+    def _registry_sync_errors(self):
```
simahawk, line 106: I would not bind this to the sync. This check should happen right after save.
What I would do:
1. add a boolean field `code_snippet_is_valid`, default=True, computed
2. add a text field `code_snippet_errors`, computed, renders the errors on the form in an alert box of type error
3. prevent the sync if invalid

The alternative is to use a constraint, which is probably better: I don't see any reason to save an endpoint that is invalid. And that's the same done by ir_actions* with `_check_python_code`.

endpoint/models/endpoint_mixin.py
```diff
@@ -98,6 +103,62 @@ def _validate_exec__code(self):
...
+            )
+        return errors
+
+    def _code_snippet_system_variable_names(self):
+        # Derive names from the actual evaluation context so this validation
+        # stays accurate when another system variable is added.
+        return set(self._get_code_snippet_eval_context(request=None))
+
+    def _code_snippet_unavailable_names(self, snippet):
+        """Find global names that safe_eval will not provide at runtime."""
```
simahawk, line 136: not sure to understand why we need this

endpoint_route_handler/models/endpoint_route_handler.py
```diff
@@ -187,6 +189,28 @@ def _check_route(self):
                     )
                 )
 
+    def _registry_sync_errors(self):
+        errors = super()._registry_sync_errors()
+        try:
+            # Binding the rule performs the same converter lookup as Odoo's
+            # routing map without modifying the live endpoint registry.
+            routing_map = werkzeug.routing.Map(
```
simahawk, line 197: To my knowledge dynamic rules are not supported by default, at least in the context of `endpoint` as they are not passed over to the snippet. What is the use case you are trying to cover?

simahawk (changes requested):

PR #160, opened 2026-08-29
[19.0][FIX] endpoint: serialize payload values with Odoo's json_default

yostashiro:
_make_json_response called json.dumps() with no default= hook, so any payload value json cannot represent natively raised TypeError and the request returned a 500. Every other JSON response in Odoo goes through json.dumps(data, default=json_default) (odoo/http.py).

Pass the same hook here. It covers date, datetime, bytes, Domain, lazy and ReadonlyDict, and json.dumps applies it at every depth, so nested values are handled without walking the payload.

Also let an endpoint supply its own hook through the result dict, for exec modes that need to override how a type is rendered. json.dumps takes a single default= callable, so such a hook replaces Odoo's for the whole payload and is expected to delegate to json_default for the types it does not render itself.

Assisted-by: Claude Opus 5

@qrtl QT6815

### DIFF
--- a/endpoint/README.rst
+++ b/endpoint/README.rst
@@ -218,6 +218,9 @@ Contributors
 - Simone Orsi <simone.orsi@camptocamp.com>
 - Iván Todorovich <ivan.todorovich@camptocamp.com>
 - Alex Garcia <alex@studio73.es>
+- `Quartile <https://www.quartile.co>`__:
+
+  - Yoshi Tashiro
 
 Maintainers
 -----------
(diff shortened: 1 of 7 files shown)


AungKoKoLin1997 (approved):
LGTM

simahawk (approved):
LGTM, thanks!
Could you backport it to v18 pls?

yostashiro:
> LGTM, thanks! Could you backport it to v18 pls?

@AungKoKoLin1997 Can you please help take care of this.

PR #161, opened 2026-08-31
[18.0][FIX] webservice: reset oauth2_flow regardless of server_environment

simahawk:
#146 split server_environment out of `webservice` into the optional `webservice_server_env` module, but the logic resetting `oauth2_flow` when `auth_type` is no longer "oauth2" only lived in `webservice_server_env`'s `_compute_server_env` override.

Without `server_environment` installed, switching a backend's `auth_type` away from "oauth2" (via the UI or a plain write) silently left a stale `oauth2_flow`, which `_get_adapter_protocol` would then still factor into the selected component.

Add an `_onchange_auth_type` plus `create`/`write` overrides on `webservice.backend` itself so the reset is guaranteed unconditionally, independent of `server_environment`. `webservice_server_env`'s own override is unchanged and still covers its own case (env-var-driven values).

CC @yankinmax @gurneyalex

### DIFF
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -100,6 +100,30 @@ class WebserviceBackend(models.Model):
         extra_params = ("auth_type",)
         return name in extra_params or super()._valid_field_parameter(field, name)
 
+    @api.onchange("auth_type")
+    def _onchange_auth_type(self):
+        # Keep `oauth2_flow` in sync in the UI as the user edits `auth_type`,
+        # regardless of whether `server_environment` is installed (see
+        # `create`/`write` below for the same guarantee on any other write).
+        if self.auth_type != "oauth2":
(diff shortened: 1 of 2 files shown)


yankinmax (approved):
I see, thank you @simahawk 
Just a question: will we FW and BKP this commit as a separate PR or do you want me just to include it in existing PR's splitting the `server_environment`?

simahawk:
FTR build red for unrelated reason https://github.com/OCA/web-api/pull/162

simahawk:
> I see, thank you @simahawk Just a question: will we FW and BKP this commit as a separate PR or do you want me just to include it in existing PR's splitting the `server_environment`?

You can already include the commit in your pending PRs.

yankinmax:
> You can already include the commit in your pending PRs.

ok, I'll include it

PR #163, opened 2026-09-02
[18.0][FIX] endpoint: serialize payload values with Odoo's json_default

AungKoKoLin1997:
Backport of #160 to 18.0.

_make_json_response called json.dumps() with no default= hook, so any payload value json cannot represent natively raised TypeError and the request returned a 500. Every other JSON response in Odoo goes through json.dumps(data, default=json_default) (odoo/http.py).

Pass the same hook here. It covers date, datetime, bytes, lazy and ReadonlyDict, and json.dumps applies it at every depth, so nested values are handled without walking the payload.

Also let an endpoint supply its own hook through the result dict, for exec modes that need to override how a type is rendered. json.dumps takes a single default= callable, so such a hook replaces Odoo's for the whole payload and is expected to delegate to json_default for the types it does not render itself.

Differences from the 19.0 PR:
- The demo endpoint covering the case is added to `demo/endpoint_demo.xml`, since 18.0 still keeps the demo endpoints in demo data rather than in `tests/common.py`.
- `Domain` does not exist in 18.0 (`ir.rule._compute_domain` returns a plain list), so the demo payload and the HTTP test only cover `date`/`datetime`.

Assisted-by: Claude Opus 5

@qrtl QT6815

### DIFF
--- a/endpoint/README.rst
+++ b/endpoint/README.rst
@@ -82,6 +82,9 @@ Contributors
 ------------
 
 - Simone Orsi <simone.orsi@camptocamp.com>
+- `Quartile <https://www.quartile.co>`__:
+
+  - Yoshi Tashiro
 
 Maintainers
 -----------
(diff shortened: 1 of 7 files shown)


yostashiro (approved):
LGTM.

simahawk (approved):

PR #166, opened 2026-09-16
[19.0][IMP] endpoint_route_handler: restore _init_modules even when a test raises

AungKoKoLin1997:
Follow-up to 369de3f, which restores `registry._init_modules` after `_get_mocked_request` but does the restore after the `with MockRequest(...)` block, without a `try/finally`. If the test body raises inside the mocked request, the exception propagates out of the generator and the restore line is never reached, so the process-wide registry is left with an empty module set again.

That attribute is not rolled back with the test transaction, and `ir.http.routing_map()` builds the routing map from exactly that set:

```python
installed = registry._init_modules.union(config["server_wide_modules"])
```

so the next routing-map regeneration yields no module controllers at all, and every later `HttpCase` in the same process gets a 404 for any route outside `base` — the very failure 369de3f set out to fix, just on the error path.

This PR wraps the `yield` in `try/finally` so the restore always runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yostashiro:
@AungKoKoLin1997 Please rebase. I guess it supplements https://github.com/OCA/web-api/pull/150 which was just merged.

yostashiro (approved):
LGTM.

simahawk:
/ocabot merge patch
Would you mind porting the whole fix to 18?

PR #167, opened 2026-09-22
[18.0][FIX] webservice: always return the full response, drop content_only

### LINKED ISSUE #142: Open Concern about webservice module defaulting content_only to True in "base.requests"

simahawk:
`_request()` defaulted `content_only` to `True`, silently returning raw
bytes instead of the `requests.Response` object callers expect. The
OAuth2 backend adapter ignored the flag entirely, always returning
bytes (fixes https://github.com/OCA/web-api/issues/142).

https://github.com/OCA/web-api/pull/144 fixes the OAuth2 inconsistency but keeps `content_only` as a
kwarg (default `None`, warns, still defaults to `True`) - perpetuating
the actual problem: the return type shouldn't be a per-call switch,
especially when raw content is just `.response.content` away.

Here, `_request()` always returns the full response; the kwarg is
dropped (a stray one is popped and warned about, not left to crash
inside `requests.request()`). Backward compat is handled once, at the
database level: the `18.0.2.0.1` upgrade script sets a
`webservice.request_content_only` system parameter on existing
databases only, preserving old behavior for calling code that relied on
the implicit default. Delete the parameter to switch to the new
behavior.

The compat switch lives in `webservice`, not `webservice_core`: the
latter is new and unreleased, so its `_request()` stays simple with no
compat baggage.


webservice/__manifest__.py
Ricardoalso: This more than a fix isn't it ? 

The module's documentation should also be adapted to reflect the new design

simahawk: done, thanks!

Ricardoalso (approved):
Thanks !

